PR #33, opened 2020-03-03
feat(examples): updates to hello-world example and directory reorganization

campionfellin:
We want more examples (as per https://github.com/awslabs/cdk8s/issues/22) but we should improve the current examples first so they work out of the box (`git clone` straight to example).

🎉First PR 🎉


campionfellin:
Looks like the `yarn build` from Lerna is failing because the `build` script in the example doesn't do the generate first.

Do we want:

1) Lerna to ignore the `/examples`
2) Examples to contain the `cdk8s generate` in build script?

I'd lean more towards 1, as I don't think `cdk8s generate` is needed for every build.

eladb:
@campionfellin I think we should reorganize `examples`  such that each example is a separate project (like yours) and change lerna and yarn from looking under `examples` to `examples/*` (should be both in `package.json` and `lerna.json`).

examples/podinfo/examples/app-example.ts
```diff
@@ -12,7 +12,7 @@ class MyChart extends Chart {
       faultError: true
     });
 
-    const deployment = new Deployment(this, 'Deployment', {
+    const deployment = new Deployment(this, 'deployment', {
```
campionfellin: I found that upper case letters were not allowed:

```
throw new Error(`"${comp}" is not a valid object name. The characters allowed in names are: digits (0-9), lower case letters (a-z), -, and .`);
                ^

Error: "Deployment" is not a valid object name. The characters allowed in names are: digits (0-9), lower case letters (a-z), -, and .
```

examples/podinfo/examples/podinfo.ts
```diff
@@ -1,5 +1,5 @@
 import { Deployment, ImagePullPolicy, IContainer } from "../lib";
-import { Container } from '../../imports/k8s';
+import { Container } from '../.gen/apps-deployment-v1'
```
campionfellin: New way of importing was not working for me. Does https://github.com/awslabs/cdk8s/commit/2db4cfbb20e7b6d8b09e8b8c9e6d096f0ea9fefe need to be published?

eladb: Yes, try to install the @next dist tag:

“yarn global add cdk8s-cli@next”
“yarn add cdk8s@next”

campionfellin: Awesome, that worked!

PR #4, opened 2020-02-17
build: add Bazel build system

CaerusKaru:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #45, opened 2020-03-05
refactor: use synchronous file system APIs and other nits

CaerusKaru:
* Move from async to sync file operations. This is done because
for a small set of files/operations, sync is actually faster than
async in Node. Also the `exists` API is deprecated, but the
`existsSync` API is not
* Move to single-quote imports where needed

### DIFF
--- a/examples/podinfo/examples/app-example.ts
+++ b/examples/podinfo/examples/app-example.ts
@@ -1,7 +1,7 @@
-import { Construct } from "@aws-cdk/core";
-import { App, Chart } from "cdk8s";
-import { Deployment, Service } from "../lib";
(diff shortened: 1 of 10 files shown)


eladb:
These changes are going to conflict with some changes I have pending and I also disagree that this is a better approach. Closing for now.

eladb:
I appreciate the desire to contribute but since this project is currently in great flux, I suggest that you raise an issue to discuss potential contributions before spending time in vane

CaerusKaru:
@eladb Thanks for the suggestion, will do in the future. For this it seemed like a trivial enough change, and I'm ok with you rejecting it outright if you don't agree. It was just a thought that came to me while working through #4

PR #46, opened 2020-03-05
refactor: remove extraneous JS file for CLI invocation

CaerusKaru:
We can specify that the bin command should resolve to a specific
JS file, which will work in a backwards-compatible way with
current invocation. This saves us having to add an extra file and
is clearer in terms of which file actually has the code being
executed

### DIFF
--- a/packages/cdk8s-cli/bin/cdk8s
+++ b/packages/cdk8s-cli/bin/cdk8s
@@ -1,2 +1,0 @@
-#!/usr/bin/env node
-require('./cdk8s.js');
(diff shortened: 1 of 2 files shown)


CaerusKaru:
I'm ok closing this for now, but I think this only fails due to how Yarn workspaces configures the bin for subpackages.

eladb:
What’s the failure?

CaerusKaru:
```bash
lerna ERR! yarn run build stderr:
/bin/sh: cdk8s: command not found
```

[Build link](https://github.com/awslabs/cdk8s/runs/488280406)

CaerusKaru:
This works in Angular, for instance, which doesn't use Yarn workspaces, see [here](https://github.com/angular/angular/blob/master/packages/compiler-cli/package.json#L10)

PR #54, opened 2020-03-08
build: add Bazel build system

CaerusKaru:
This is a continuation of #4, which got a little messed up when the repo was made public.


CaerusKaru:
Apologies to anyone tracking for the commit spam. Here's the current status:

- All build steps pass locally and on CI. This includes all packaging and release steps through jsii-release
- All tests currently build and run correctly. The only errors are due to out-of-date snapshots. Functionality to update the snapshots through Bazel is pending
- All examples build and generate the imports files through Bazel correctly, in addition to tests. However example synthesis is still pending (though is mostly done). Though strictly speaking they synthesize correctly, they don't end up in the correct directory for easy use.

tl;dr only two things left: snapshot test updation through Bazel, example synthesis through Bazel

CaerusKaru:
Progress update:

- Jest tests all pass and snapshot updation is covered and working. The CLI tests need an NPM patch so that `require.resolve` will work correctly with JSII
- Example synthesis is still pending, but is the only major hurdle left

.github/.bazelrc
```diff
@@ -0,0 +1,12 @@
+# Bazel configuration that will be copied to /etc/bazel.bazelrc.
```
eladb: What is this file? Why is it under `.github/`? How is it different from the root `.bazelrc`?

.github/workflows/build.yml
```diff
@@ -10,6 +10,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: yarn install
+      - run: yarn bazelrc:ci
+      - run: yarn bazelrc
```
eladb: What are these doing?

BUILD.bazel
```diff
@@ -0,0 +1,5 @@
+package(default_visibility = ["//visibility:public"])
+
+exports_files([
+    "LICENSE",
+])
```
eladb: What does this do? Why is it needed here?

BUILD.bazel
```diff
@@ -0,0 +1,5 @@
+package(default_visibility = ["//visibility:public"])
```
eladb: What is the meaning of the `BUILD.bazel` file at the root? Does it represent the entire project?

WORKSPACE
```diff
@@ -0,0 +1,64 @@
+workspace(
+    name = "cdk8s_srcs",
+    managed_directories = {"@npm": ["node_modules"]},
+)
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
```
eladb: what is skylib? add a comment explaining why is it needed?

examples/hello/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
+package(default_visibility = ["//visibility:public"])
+
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
+load("//tools:cdk8s.bzl", "cdk8s_import", "cdk8s_synth")
+load("//tools:defaults.bzl", "jest_test", "ts_library")
+
+cdk8s_import(
+    name = "gen",
```
eladb:
```suggestion
    name = "import",
```

examples/hello/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
...
+    path = "examples/hello/imports",
```
eladb: feels like this should be a relative path, and I we should also make `./imports` the default, so basically I'd expect this to be omitted.

examples/hello/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
...
+        exclude = ["**/*.test.ts"],
+    ),
+    deps = [
+        "//packages/cdk8s:lib",
+        "@npm//@aws-cdk/core",
+    ],
+)
+
+filegroup(
+    name = "sources",
```
eladb: feels like this should be reused in the `ts_library` invocation above, no? I wouldn't want to specify sources twice

examples/hello/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
+package(default_visibility = ["//visibility:public"])
```
eladb: generally, this feels wayyy to long... All examples will have roughly the same build definition. Can we generalize this somehow (i.e. create something like `cdk8s_example(...)` command?). I feel this file should be a single line with all defaults.

examples/hello/jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
eladb: This shouldn't be duplicated in all examples, I expect to be centrally managed

examples/hello/package.json
```diff
@@ -5,11 +5,6 @@
   "types": "index.d.ts",
```
eladb: didn't we say we wanted package.json files to be generated?

examples/hello/snapshotResolver.js
```diff
@@ -0,0 +1,15 @@
+const {basename, dirname, join} = require('path');
```
eladb: what the ?? why is this needed and how can we get rid of this?

examples/podinfo/BUILD.bazel
```diff
@@ -0,0 +1,64 @@
...
+    testonly = True,
+    srcs = [":lib"],
+    output_group = "es5_sources",
+)
+
+jest_test(
+    name = "test",
+    srcs = glob(["**/*.snap"]) + [
+        ":test_sources",
+        ":sources",
```
eladb: same note, this whole file needs to be a single line

package.json
```diff
@@ -3,26 +3,58 @@
   "version": "0.12.0",
   "private": true,
   "scripts": {
+    "bazel": "bazelisk",
+    "bazelrc:ci": "cp ./.github/.bazelrc /etc/bazel.bazelrc",
+    "bazelrc": "tools/generate-bazelrc.sh",
+    "bazel:buildifier": "find . -type f \\( -name \"*.bzl\" -or -name WORKSPACE -or -name BUILD -or -name BUILD.bazel \\) ! -path \"*/node_modules/*\" | xargs buildifier -v --warnings=attr-cfg,attr-license,attr-non-empty,attr-output-default,attr-single-file,constant-glob,ctx-args,depset-iteration,depset-union,dict-concatenation,duplicated-name,filetype,git-repository,http-archive,integer-division,load,load-on-top,native-build,native-package,output-group,package-name,package-on-top,redefined-variable,repository-name,same-origin-load,string-iteration,unused-variable,unsorted-dict-items,out-of-order-load",
```
eladb: wtf?

packages/README.md
```diff
@@ -0,0 +1,6 @@
+AWS CDK8s
```
eladb: why do we need a readme file here?

packages/BUILD.bazel
```diff
@@ -0,0 +1,6 @@
+package(default_visibility = ["//visibility:public"])
+
+exports_files([
+    "tsconfig-build.json",
```
eladb: ??

packages/cdk8s-cli/BUILD.bazel
```diff
@@ -0,0 +1,116 @@
...
+    name = "package_json",
+    package_name = "cdk8s-cli",
+    bin = {
+        "cdk8s": "bin/cdk8s.js",
+    },
+    desc = "CDK for Kubernetes CLI",
+    experimental = True,
+    peer_deps = {
+        "codemaker": "^0.22.0",
+        "yargs": "^15.1.0",
```
eladb: I don't think these should be peer deps. I think it was a mistake in our original package.json

packages/cdk8s-cli/BUILD.bazel
```diff
@@ -0,0 +1,116 @@
...
+)
+
+filegroup(
+    name = "test_sources",
+    testonly = True,
+    srcs = [":tests"],
+    output_group = "es5_sources",
+)
+
+filegroup(
```
eladb: again feels like this is duplicated with how all the examples are configured. definitely seems like an opportunity for reuse

packages/cdk8s-cli/jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  "collectCoverage": true,
+  "moduleFileExtensions": [
+    "js"
```
eladb: omit

packages/cdk8s-cli/snapshotResolver.js
```diff
@@ -0,0 +1,15 @@
+const {basename, dirname, join} = require('path');
+const bazelDir = process.env['BUILD_WORKSPACE_DIRECTORY'] || process.cwd();
+const bazelRoot = process.env['BAZEL_PATCH_ROOT'] || '';
+const binDir = 'bin';
+const pathEnd = dirname(bazelRoot.slice(bazelRoot.lastIndexOf(binDir) + binDir.length));
```
eladb: HAA?A?

.github/.bazelrc
CaerusKaru: The root `.bazelrc` is for all environments and users. This file is only used under CI because it has specific commands that only CI should use (as mentioned below). If you follow the link, you'll see that there's precedence given to Bazel RC files in the loading phase:

- Top-level (/etc/bazel.bazelrc)
- Workspace-level (.bazelrc)
- User-level (.bazelrc.user - loaded by .bazelrc at the end of the file)

.github/workflows/build.yml
CaerusKaru: In order for Bazel to get setup correctly, we need to move the CI bazelrc file into place (`/etc/bazel.bazelrc`) and then we need to initialize Bazel with the CI's PATH variables (any any other ENV we need at build time). I can add comments to this file if you think it will help.

BUILD.bazel
CaerusKaru: The LICENSE file gets copied from the root into each of the underlying packages and build time. This rule is needed to make the LICENSE file available to subpackage BUILD rules.

BUILD.bazel
CaerusKaru: It represents all targets at the root of the project. In this case, specifically we use it to create a target for the LICENSE file that we can consume in other rules.

WORKSPACE
CaerusKaru: This is, in fact, not needed. At one point we depended on it for the `bzl_library` rule, but I got rid of that in one of the drafts. Will remove in the next commit.

examples/hello/BUILD.bazel
CaerusKaru: Yes we can, I'll create a macro for this in the next commit. I had this inspiration last night too, but you beat me to it 😉

examples/hello/BUILD.bazel
CaerusKaru: I'll have to check to see if we can do that, but the reason is that we need to provide a relative path so that the CLI knows where to output to relative to the workspace root. Again, we may be able to get around this, but I'll have to double check.

examples/hello/BUILD.bazel
CaerusKaru: The name sources here might be a little misleading. In this case what this means is the "compiled JS source" _from_ the `ts_library` rule. I can add a comment if this is still unclear.

examples/hello/jest.config.js
CaerusKaru: The reason this isn't centrally managed vis-a-vis the package tests is because the examples maintain a different structure for where snapshots are located, so they need a separate snapshot resolver. If you want, we can change the directory structure to be dogmatic, but if you want to keep this structure, we need a separate config here.

examples/hello/package.json
CaerusKaru: I kept this and the other example package.json files in case you wanted them available for users to take as examples as well. I agree it doesn't make sense project-wise to keep them, and we don't actually use them in any real way here. I can get rid of them if you want, but I didn't want to delete them unless you were sure.

examples/hello/snapshotResolver.js
CaerusKaru: All of the Bazel tests run in a sandbox (at least on UNIX). So although _using_ snapshots is totally covered since we just copy over the snapshots into the sandbox, _updating_ snapshots is a totally different story. We need to be able to tell Jest where to save the snapshots. This feature was introduced to Jest explicitly to cover the Bazel use case (see original PR [here](https://github.com/facebook/jest/pull/6143)). While we can centralize it, if you want to maintain the functionality of keeping snapshots up-to-date with Bazel, we need this.

The alternative is to install ts-jest that can run unit tests on the TS source outside of Bazel and then update the snapshots that way, but believe me that this is cleaner and more straightforward. Once you get over the sticker shock of this file, it makes a lot of sense and eases a lot of pain.

examples/podinfo/BUILD.bazel
CaerusKaru: Agreed, will add macro.

package.json
CaerusKaru: This is basically one giant lint rule for Bazel BUILD/WORKSPACE files. I agree it's lengthy, and I'll talk to the team about finding a way to consolidate it, but it works well and covers all of the basic needs of a linter.

packages/BUILD.bazel
CaerusKaru: We need a way for `ts_library` to centrally consume these files, so we export them from here. We can move this to the `tools` directory, but traditionally it's here to aid in discoverability (i.e. the tsconfig lives in the same part of the package as the packages it configures)

packages/README.md
CaerusKaru: Strictly speaking we don't. Traditionally, it's here as a fallback in case a package doesn't have a README of it's own, then this gets plugged in. But I can get rid of it since it doesn't look like we need that functionality.

packages/cdk8s-cli/BUILD.bazel
CaerusKaru: Sounds like I missed that in one of the rebases. I'll remove based on the current spec.

packages/cdk8s-cli/BUILD.bazel
CaerusKaru: Agreed, once all functionality is up and running, consolidation is job one 😄

packages/cdk8s-cli/jest.config.js
CaerusKaru: Do you mean centralize, or can we get rid of the `moduleFileExtensions` rule?

packages/cdk8s-cli/snapshotResolver.js
CaerusKaru: I'll add a comment to the header of this file, but basically we need to pierce of the veil of the Bazel sandbox in order to update snapshots. Otherwise the updated snapshots would just exist in the sandbox and never get moved back to the source. See earlier comment about why this functionality was added to Jest in the first place.

WORKSPACE
CaerusKaru: Redacted above. We do need this for `copy_bash` used in the `jsii_library`. I'll add a comment.

.github/.bazelrc
eladb: Let's move it to the root and rename it `.bazelrc-ci` or something like that so it will be discoverable next to the root one.

.github/workflows/build.yml
eladb: Why do we need two commands here?

WORKSPACE
eladb: Wouldn't it make sense to load it as part of the definition of `jsii_library` instead of at the workspace level?

examples/hello/jest.config.js
eladb: Still not sure why we can't reuse this file across all examples?

examples/hello/package.json
eladb: Let's get rid of them.

examples/hello/snapshotResolver.js
eladb: We can't have this file needed in all examples. I actually wanted to move to `ts-jest` anyway, so let's do that.

package.json
eladb: can't we put it in under `tools/bazel-buildiufer.sh` or something?

packages/README.md
eladb: LEt's

packages/BUILD.bazel
eladb: Since we have a central tsconfig file, I would expect this to be centralized as well and not under packages/

packages/cdk8s-cli/jest.config.js
eladb: yes, centralize. I don't want any duplication between packages... I see this as an important benefit of Bazel that we should double down on.

packages/cdk8s-cli/snapshotResolver.js
eladb: ts-jest you said, no?

.github/.bazelrc
CaerusKaru: When you say this, do you mean instead of copying to `/etc/bazel.bazelrc`, or just that it shouldn't be hidden in `.github` to start?

WORKSPACE
CaerusKaru: It's loaded there, but it's _downloaded_ here. That's how Bazel works, all necessary tools/libraries are fetched and setup in the WORKSPACE file, then consumed in the BUILD/*.bzl files.

packages/BUILD.bazel
CaerusKaru: The root tsconfig.json is primarily for IDE consumption, but we can tweak it to be reused, sure.

.github/.bazelrc
eladb: Shouldn't be hidden

.github/workflows/build.yml
CaerusKaru: The first one moves the top-level Bazel RC file into place, only on CI. The second is a command used by all consumers of the library. We could fold it into `prebuild` (I have a comment on that in the package.json right now), but both commands are required. We could rename them too to make it clearer.

CaerusKaru:
Progress update:

- CLI tests are working fine now. Snapshot updation has a minor issue in the examples, but tests run regardless
- Examples synth is still pending
- Fixed a packaging issue that preventing `yaml` from getting bundled with the library. Also added the LICENSE file

packages/BUILD.bazel
CaerusKaru: So we can't consolidate, because the root tsconfig depends on having `includes`, `outDir`, and `paths`, whereas the build tsconfig does not. Consolidating them breaks the build. Fortunately though we only have one tsconfig for every part of the build library.

examples/hello/BUILD.bazel
CaerusKaru: Turns out we can do it with a bit of bash magic 😄

CaerusKaru:
All systems go! There is one precondition for doing example synthesis that I'm trying to make easier, but in the meantime the base functionality for the entire repo has been replicated with Bazel in some form.

eladb:
@CaerusKaru can you do one last merge from master and I'll pick this up

CaerusKaru:
@eladb Should be all set :+1:

eladb:
I am getting the following message when running "yarn install":

```
Unrecognized patch file in patches directory rules_nodejs.patch
```

packages/cdk8s/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+
+package_json(
+    name = "package_json",
+    package_name = MODULE_NAME,
+    bundled_deps = ["json-stable-stringify", "yaml"],
+    desc = "Cloud Development Kit for Kubernetes",
+    experimental = True,
+    jsii_targets = JSII_TARGETS,
+    peer_deps = {
+        "constructs": CONSTRUCTS_PACKAGE_VERSION,
```
eladb: I expected `constructs` to also appear in `devDeps` here. I actually want to make sure the `devDep` is pinned on 2.0.0 because that's the lowest version we currently support.

packages/cdk8s/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+        "package": "org.cdk8s",
+    },
+    "python": {
+        "distName": "cdk8s",
+        "module": "cdk8s",
+    },
+})
+
+package_json(
+    name = "package_json",
```
eladb: Is this `name` really needed? This produces a package.json file, can we have a default on the name?

packages/cdk8s/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+    package_name = MODULE_NAME,
```
eladb: seems inconsistent. In some places this is called `module_name` and here `package_name`. Can we normalize?

packages/cdk8s/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+)
+
+jsii_library(
+    name = "lib",
+    srcs = glob(
+        ["**/*.ts"],
+        exclude = ["test/**"],
+    ),
+    module_name = MODULE_NAME,
+    package_json = ":package_json",
```
eladb: Can probably be omitted if we have a default name for the `package_json` rule, no?

packages/cdk8s/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+    deps = [
+        "@npm//@types/json-stable-stringify",
+        "@npm//@types/node",
+        "@npm//@types/yaml",
+        "@npm//constructs",
```
eladb: Feels like a duplicate from package.json, can we remove?

packages/cdk8s/BUILD.bazel
eladb: why isn't this @npm//constructs or something like that?

packages/cdk8s/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+    bundled_deps = ["json-stable-stringify", "yaml"],
+    desc = "Cloud Development Kit for Kubernetes",
+    experimental = True,
+    jsii_targets = JSII_TARGETS,
+    peer_deps = {
+        "constructs": CONSTRUCTS_PACKAGE_VERSION,
+    },
+    deps = {
+        "json-stable-stringify": "^1.0.1",
+        "yaml": "^1.7.2",
```
eladb: I expected these to reference the @npm// scope somehow instead of hard code the versions here, no?

packages/cdk8s/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+    deps = [
+        ":lib",
+        "@npm//constructs",
+        "@npm//json-stable-stringify",
+        "@npm//tslib",
+        "@npm//yaml",
+    ],
+)
+
+jsii_package(
```
eladb: What's the difference between `jsii_library` and `jsii_package`? Why do we need both?

packages/cdk8s/BUILD.bazel
```diff
@@ -0,0 +1,104 @@
...
+        },
+        "package": "org.cdk8s",
+    },
+    "python": {
+        "distName": "cdk8s",
+        "module": "cdk8s",
+    },
+})
+
+package_json(
```
eladb: I am wondering if we even need `package_json` here. Seems like we can probably merge it (and `jsii_package`) into `jsii_library` which will just do everything. I don't see any reason to split those up really.

Basically, I'd expect this file to look something like this:

```bzl
jsii_library(
    name = "lib",
    desc = "Cloud Development Kit for Kubernetes",
    stability = "experimental",
    jsii_targets = JSII_TARGETS,
    srcs = glob(
        [ "**/*.ts", "README.md" ],
        exclude = ["test/**"],
    ),
    module_name = "cdk8s",
    dev_deps = [
        "@npm//@types/json-stable-stringify",
        "@npm//@types/node",
        "@npm//@types/yaml",
        "@npm//constructs", # this needs to be pinned to 2.0.0
    ],
    peer_deps = [
        "@npm//constructs",  # this needs to be ^2.0.0 (not sure how to express)
    ],
    bundled_deps = [
        "json-stable-stringify",
        "yaml"
    ]
)
```

Something along those lines.

eladb:
@CaerusKaru this is not good enough yet... too much duplicated information all over the place. Repeating that one of the main motivations for me is to "DRY"-up all the info that we need to supply in order to build. I want to be able to provide the minimum amount of information and never repeat myself if not needed.

packages/cdk8s/BUILD.bazel
CaerusKaru: I think it's important to have separate, same with `jsii_package`, because it's a separate action. We can create a macro to encapsulate the behavior you're talking about, but leaving these compose-able gives us some advantages. E.g. we don't need to recompile a library every time we want to package it. Or we can reuse the package.json elsewhere without compiling the library (important for test path resolution).

packages/cdk8s/BUILD.bazel
CaerusKaru: We can have a macro that defaults the name, sure. But the name corresponds to the target and is a mandatory parameter for Bazel actions.

packages/cdk8s/BUILD.bazel
CaerusKaru: `module_name` is a `ts_library` parameter that we inherit (this allows us to use `jsii_library` as a dep for `ts_library` targets). `package_name` can be converted to `module_name` for consistency, but I thought `package_name` was clearer here.

packages/cdk8s/BUILD.bazel
CaerusKaru: `devDeps` are not needed for package publishing, otherwise package version is handled solely at the root package.json level.

CaerusKaru: This is translated raw to the package.json. It doesn't incorporate the actual package as a dep (though it could if we decide that's the way to go).

packages/cdk8s/BUILD.bazel
CaerusKaru: That could be incorporated into the package.json rule, sure.

packages/cdk8s/BUILD.bazel
CaerusKaru: Only if we also adapt the `jsii_library` to be a macro that accepts the default. In my mind, we'd have one `jsii` macro that sets up the `package_json`, `jsii_library` and `jsii_package` targets all in one. This has the benefit of being DRY but also allowing the user to trigger the individual actions.

packages/cdk8s/BUILD.bazel
CaerusKaru: Only if we incorporate the deps aspect into the package_json rule, but sure.

packages/cdk8s/BUILD.bazel
CaerusKaru: `jsii_library` -> `jsii`
`jsii_package` -> `jsii-pacmak`

Each requires separate tooling, and so should remain separate (when you develop a project, you don't invoke `jsii-pacmak` every time)

CaerusKaru:
@eladb There is more we can do in terms of compartmentalizing by creating macros, but really it's up to you to determine the ergonomics you want. So far my goal has just been to get the project "working" in an MVP state. How you want it to work after that is really just gravy now that it does work.

eladb:
Yes I understand. Let me try to play with this a little to see if I can get to the ergonomics I want

eladb:
@CaerusKaru I think I am going to close this one for the time being. I don't think I have the time in the next few months to spend on getting this to a place that I would feel comfortable merging. @iliapolo would you be interested to look into this perhaps?

eladb:
@CaerusKaru I really appreciate all the effort!

iliapolo:
@el

iliapolo:
@eladb definitely. I’ll see if I can pick this up.

PR #67, opened 2020-03-12
feat: sort keys of ApiObject manifests

### LINKED ISSUE #17: [Suggestion] fix ordering in YAML output
**Is your feature request related to a problem? Please describe.**
Synthesized YAML shows up out of order from standard K8s format. It technically works, but for a user familiar with K8s YAML formatting can be confusing to read.

example:
```
spec:
  type: LoadBalancer
  ports:
    - port: 80
      targetPort: 8080
  selector:
    app: myappnginx37F28063
kind: Service
apiVersion: v1
metadata:
  name: myapp.nginx.service.5f289726
```

**Describe the solution you'd like**
Follow K8s YAML formatting conventions
1. apiVersion
2. kind
3. metadata
4. spec
5. etc...

example:
```
apiVersion: v1
kind: Service
metadata:
  name: myapp.nginx.service.5f289726
spec:
  type: LoadBalancer
  ports:
    - port: 80
      targetPort: 8080
  selector:
    app: myappnginx37F28063
```

_There may be a styling guide somewhere but I can't find it - looking at [K8s documentation](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service)_

campionfellin:
Fixes #17 

Using [`json-stable-stringify`](https://github.com/substack/json-stable-stringify) to sort output alphabetically and recursively. 

### DIFF
--- a/packages/cdk8s-cli/test/__snapshots__/integration.test.js.snap
+++ b/packages/cdk8s-cli/test/__snapshots__/integration.test.js.snap
@@ -87,9 +87,9 @@ Object {
     },
   },
   "description": "dummy",
-  "fingerprint": "mdQ/Dby2xkDPeGSW31K8SNn/rUvqDSFgyzbNqu5N96c=",
+  "fingerprint": "x9MKEvQhXUWZn4wUbWuapgOUz/7jFDJHiOKmDGti2Wg=",
   "homepage": "http://repo",
-  "jsiiVersion": "1.1.0 (build df55f5e)",
+  "jsiiVersion": "0.22.0 (build 14afdde)",
   "license": "Apache-2.0",
   "name": "dummy",
   "repository": Object {
--- a/packages/cdk8s-cli/test/__snapshots__/integration.test.js.snap
+++ b/packages/cdk8s-cli/test/__snapshots__/integration.test.js.snap
@@ -87,9 +87,9 @@ Object {
     },
   },
   "description": "dummy",
-  "fingerprint": "HD4O8JPi7/yzcp+ASh//v9CjPwF8WUUjrfUevSbCO9A=",
+  "fingerprint": "mdQ/Dby2xkDPeGSW31K8SNn/rUvqDSFgyzbNqu5N96c=",
   "homepage": "http://repo",
-  "jsiiVersion": "1.0.0 (build eb2db9d)",
+  "jsiiVersion": "1.1.0 (build df55f5e)",
   "license": "Apache-2.0",
   "name": "dummy",
   "repository": Object {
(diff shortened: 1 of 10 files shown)


packages/cdk8s/lib/api-object.ts
```diff
@@ -121,7 +122,7 @@ export class ApiObject extends Construct {
 
     // convert to "pure data" so, for example, when we convert to yaml these
     // references are not converted to anchors.
-    return JSON.parse(JSON.stringify(removeEmpty(resolve(this, data))));
+    return JSON.parse(stringify(removeEmpty(resolve(this, data))));
```
campionfellin: Doing this at the `apiObject` level instead of `Chart` level. I think it makes more sense to do it here.

That being said... I just found that `YAML` version `1.8.0` now supports the flag `sortMapEntries` (https://github.com/eemeli/yaml/releases/tag/v1.8.0)... I'll take a look there as well

packages/cdk8s/package.json
```diff
@@ -75,6 +76,7 @@
     "jsii": "^0.22.0",
     "jsii-pacmak": "^0.22.0",
     "json-schema-to-typescript": "^8.0.1",
+    "json-stable-stringify": "^1.0.1",
```
eladb: This needs to be added as a `bundledDependency` since it's not a jsii module

eladb: You will also need to define it in the root `package.json` under `noHoist`

packages/cdk8s/lib/api-object.ts
eladb: I like it that we are doing this at the JSON-level inside `ApiObject` and not when we convert to YAML.

campionfellin: Awesome! I'l keep it here then.

packages/cdk8s-cli/test/__snapshots__/integration.test.js.snap
```diff
@@ -87,9 +87,9 @@ Object {
     },
   },
   "description": "dummy",
-  "fingerprint": "HD4O8JPi7/yzcp+ASh//v9CjPwF8WUUjrfUevSbCO9A=",
+  "fingerprint": "mdQ/Dby2xkDPeGSW31K8SNn/rUvqDSFgyzbNqu5N96c=",
   "homepage": "http://repo",
-  "jsiiVersion": "1.0.0 (build eb2db9d)",
+  "jsiiVersion": "1.1.0 (build df55f5e)",
```
eladb: Please revert to 0.22.0 (the AWS CDK still uses this version due to some issue we are having).

campionfellin: Done!

tabern:
Nice work - thanks @campionfellin

PR #87, opened 2020-03-28
fix(examples): "replicas" option is not respected in the web-service example

cmdallas:


examples/web-service/web-service.ts
```diff
@@ -31,6 +31,7 @@ export class WebService extends Construct {
     const port = options.port || 80;
     const containerPort = options.containerPort || 8080;
     const label = { app: Node.of(this).uniqueId };
+    const replicas = options.replicas || 1;
```
eladb: Use “??” Instead “||” please.

PR #101, opened 2020-04-11
feat(example): python hello example

campionfellin:
First pass at converting more-or-less the [python into](https://github.com/awslabs/cdk8s/blob/master/docs/intro-python.md)

to an actual example, that will be built by `lerna` with `yarn build`. We can use this to verify changes continue to work with python. It's also a great point of reference for python developers using `cdk8s`


examples/hello-python/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "hello-python",
+  "version": "0.0.0",
+  "license": "Apache-2.0",
+  "private": true,
+  "scripts": {
+    "build": "cdk8s import --language python",
+    "synth": "cdk8s synth",
+    "test": "Echo 'no test for python yet.'"
```
campionfellin, line 9:
```suggestion
    "test": "echo 'no test for python yet.'"
```

campionfellin:
@eladb this feels a bit odd, with `package.json` in it. Is there a different/better way to get `lerna` to build this?

eladb:
> @eladb this feels a bit odd, with `package.json` in it. Is there a different/better way to get `lerna` to build this?

You need `package.json` and that's fine. The problem is that this is going to be built against the published version of the CDK and not against the local build. To do that you'll need this to actually run *after* package, and you'll have to install the python module from `dist/python` instead of from PyPI.

Not 100% sure what's the best way to do that TBH.

campionfellin:
@eladb I looked into it, and it seems I just needed the dependency on the `cdk8s-cli`, which is supplied by `node_modules/.bin/cdk8s` (which is in-place after a `yarn install` from lerna at the root of the repository).

I think folks can take the example and edit as-needed (remove package.json, add deps to Pipefile, etc), but this stands as a base starting point for users.

campionfellin:
@eladb let's merge this and the other python examples

PR #103, opened 2020-04-12
chore: homebrew release download-url is invalid

eladb:
The default download URL compiled by the homebrew release action is `https://github.com/${owner}/${repo}/archive/${tagName}.tar.g` but our formula downloads the tarball from npm. Also, we don't trigger on a tag name anymore, so we can't use it to determine the version.

This change reads the version from `package.json` and provides a custom `download-url` to the action.

---


.github/workflows/release.yml
```diff
@@ -113,9 +113,18 @@ jobs:
...
+        run: |
+          version=$(node -p "require('./package.json').version")
+          echo "::set-output name=version::${version}"
+
+      # A PR will be sent to github.com/Homebrew/homebrew-core to update this formula:
       - uses: mislav/bump-homebrew-formula-action@v1
         with:
-          # A PR will be sent to github.com/Homebrew/homebrew-core to update this formula:
           formula-name: cdk8s
+          download-url: https://registry.npmjs.org/cdk8s-cli/-/cdk8s-cli-${{ steps.get_version.outputs.version }}.tgz
```
eladb, line 128: We need to also set a custom commit title

campionfellin:
@eladb I wonder what the tradeoff is with maintaining this release script ourselves vs just letting the `homebrew-core` folks update the version.

I'm seeing both `0.18.0` and `0.19.0` got picked up by them. I guess we'd be giving up some control, but would not have to worry about this script in the future.

eladb:
> @eladb I wonder what the tradeoff is with maintaining this release script ourselves vs just letting the `homebrew-core` folks update the version.
> 
> 
> 
> I'm seeing both `0.18.0` and `0.19.0` got picked up by them. I guess we'd be giving up some control, but would not have to worry about this script in the future.

What does that mean “picked up by them”? How do they know that there is a new release?

campionfellin:
Looking at their PRs for `cdk8s` [here](https://github.com/Homebrew/homebrew-core/pulls?q=is%3Apr+is%3Aclosed+cdk8s) it seems one of the maintainers is running this script:

```
brew bump-formula-pr.
```

You can see what this does with `brew bump-formula-pr --help` locally.

Example PR: https://github.com/Homebrew/homebrew-core/pull/52885

Not sure how often this gets run (maybe it's slightly delayed from when we actually release).

PR #108, opened 2020-04-17
fix(readme): missing information about imports in cdk8s-cli readme

campionfellin:
Updates the `cdk8s-cli` readme with some of the recent changes to importing.

Fixes: https://github.com/awslabs/cdk8s/issues/28

### DIFF
--- a/packages/cdk8s-cli/README.md
+++ b/packages/cdk8s-cli/README.md
@@ -16,7 +16,9 @@ Install globally via `yarn` (or `npm`):
 yarn global add cdk8s-cli
 ```
 
-## init - Create new projects
+## init
+
+### Create new projects
 
 This command creates new cdk8s projects from built in templates:
 
@@ -32,7 +34,9 @@ $ cd my-fun-little-project
 $ cdk8s init typescript-app
 ```
 
-## import - Import API objects as constructs
+## import
+
+### Import API objects as constructs
 
 To generate constructs for all Kubernetes API objects of a certain version, use
 the `import` subcommand:
@@ -51,6 +55,46 @@ objects:
 ```shell
 $ cdk8s import k8s
 ```
+
+### Import different versions of the k8s API
+
+By default, the command `cdk8s import k8s` will use the latest stable version of k8s objects.
+
+For example, when using k8s 1.17,  if you want to use `DeploymentSpec`, you would be importing it from [Deployment v1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#deploymentspec-v1-apps) by default. 
+
+In the default `v1`, the `revisionHistoryLimit`, which is the number of old ReplicaSets to retain, is set to 10 by default. You would have to explicitly set this to a higher number if you want to retain more.
+
+Let's say that you want the behavior of [extensions/v1beta1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#deploymentspec-v1beta1-extensions), which defaults to retaining _all_ old ReplicaSets.
+
+To do this, you could add the `--include` flag to your `cdk8s import`. Here is how that would look:
+
+```code
+cdk8s import k8s --language typescript --include io.k8s.api.extensions.v1beta1.Deployment
+```
+
+Here we are explicitly choosing to use the [Deployment extensions/v1beta1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#deployment-v1beta1-extensions) version, which includes the new default behavior that we want from `DeploymentSpec`
+
+### Import Customer Resource Definitions (CRDs)
+
+You can import CRDs from local files (`my_crd.yml`) or from a remote host (`https://raw.githubusercontent.com/jenkinsci/kubernetes-operator/master/deploy/crds/jenkins.io_jenkins_crd.yaml`)
+
+Just like other imports, this is how you do that:
+
+```code
+cdk8s import my_crd.yaml
+```
+
+By default, your imported CRD will be named by its `spec.names.kind`. If your CRD file contains multiple CRDs, this will create multiple imports, each named by its respective `spec.names.kind`.
+
+### Override default behavior of import naming
+
+You can also override the default behavior for `cdk8s` importing. This will be helpful to you if you would like to name your CRD imports differently from their kind. Here is how you do it:
+
+```code
+cdk8s import example:=my_crd.yaml
+```
+
+In this case, your import would be named `example`.
 
 ## License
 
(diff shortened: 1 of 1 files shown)


packages/cdk8s-cli/README.md
```diff
@@ -32,7 +34,9 @@ $ cd my-fun-little-project
 $ cdk8s init typescript-app
 ```
 
-## import - Import API objects as constructs
+## import
+
+### Import API objects as constructs
 
 To generate constructs for all Kubernetes API objects of a certain version, use
```
eladb, line 41: This needs an update. `import` can now also be used to generate constructs for CRDs as well

packages/cdk8s-cli/README.md
```diff
@@ -32,7 +34,9 @@ $ cd my-fun-little-project
 $ cdk8s init typescript-app
 ```
 
-## import - Import API objects as constructs
+## import
+
+### Import API objects as constructs
```
eladb, line 40: We also need to mention the syntax in `cdk8s.yaml`

packages/cdk8s-cli/README.md
```diff
@@ -52,6 +56,46 @@ objects:
 $ cdk8s import k8s
 ```
 
+### Import different versions of the k8s API
```
eladb, line 59: I think this needs to be an H4 as it is an option of `k8s`, no?

```suggestion
#### Selecting `apiVersion`s for specific objects
```

packages/cdk8s-cli/README.md
```diff
@@ -52,6 +56,46 @@ objects:
...
+Let's say that you want the behavior of [extensions/v1beta1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#deploymentspec-v1beta1-extensions), which defaults to retaining _all_ old ReplicaSets.
+
+To do this, you could add the `--include` flag to your `cdk8s import`. Here is how that would look:
+
+```code
+cdk8s import k8s --language typescript --include io.k8s.api.extensions.v1beta1.Deployment
+```
+
+Here we are explicitly choosing to use the [Deployment extensions/v1beta1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#deployment-v1beta1-extensions) version, which includes the new default behavior that we want from `DeploymentSpec`
+
```
eladb, line 76: Can you please raise an issue to add support for this in `cdk8s.yaml`?

packages/cdk8s-cli/README.md
```diff
@@ -52,6 +56,46 @@ objects:
...
+### Import Customer Resource Definitions (CRDs)
```
eladb, line 77:
```suggestion
### Import Custom Resource Definitions (CRDs)
```

packages/cdk8s-cli/README.md
```diff
@@ -52,6 +56,46 @@ objects:
...
+
+### Override default behavior of import naming
+
+You can also override the default behavior for `cdk8s` importing. This will be helpful to you if you would like to name your CRD imports differently from their kind. Here is how you do it:
+
+```code
+cdk8s import example:=my_crd.yaml
+```
+
+In this case, your import would be named `example`.
```
eladb, line 97: Not accurate, right? This will only be used as a prefix, while the suffix is still going to be the CRD kind name.
Also worth mentioning how this plays out in each language.

packages/cdk8s-cli/README.md
campionfellin: Yep! https://github.com/awslabs/cdk8s/issues/110

packages/cdk8s-cli/README.md
campionfellin: Mentioned: `Note: this is not yet supported for python.`

packages/cdk8s-cli/README.md
```diff
@@ -32,24 +34,148 @@ $ cd my-fun-little-project
...
+`cdk8s` CLI
+
+```code
+cdk8s import --language typescript
+```
+
+`cdk8s.yaml` Config
+
+```yaml
+language: typescript
```
eladb, line 52: Language is probably going to be used for other things as well so I wouldn’t put it under the imports section.

packages/cdk8s-cli/README.md
```diff
@@ -32,24 +34,148 @@ $ cd my-fun-little-project
...
+```
+imports/
+  k8s.ts
+```
+
+Example output for `python` imports is:
+
+```
+imports/
+  k8s/
```
eladb, line 68: That feels redundant. I think the import example code should be how to use the import in each language

packages/cdk8s-cli/README.md
```diff
@@ -32,24 +34,148 @@ $ cd my-fun-little-project
...
 
 ```shell
-$ cdk8s import SPEC
+$ cdk8s import k8s
 ```
 
-Currently, the only supported spec is `"k8s"` which represents the core
-Kubernetes API. The import command will create a file under `imports/k8s.ts`
-with constructs for each API object in the spec.
+By default, the import command will create a file under `imports/k8s.ts`
```
eladb, line 85: This is only true for typescript

packages/cdk8s-cli/README.md
```diff
@@ -32,24 +34,148 @@ $ cd my-fun-little-project
...
+  - my_crd.yaml
+```
+
+By default, your imported CRD will be named by its `spec.names.kind`. If your CRD file contains multiple CRDs, this will create multiple imports, each named by its respective `spec.names.kind`.
+
+For example, if `my_crd.yaml` contained two kinds, `cluster` and `autoscaler`, you would have two imports:
+
+```
+imports/
+  cluster.ts
```
eladb, line 150: Again, replace this with an examples on how to consume this import from each supported language.

packages/cdk8s-cli/README.md
```diff
@@ -32,24 +34,148 @@ $ cd my-fun-little-project
...
+```yaml
+imports:
+  - example:=my_crd.yaml
+```
+
+If your CRD contained two kinds, `cluster` and `autoscaler`, you would have two imports:
+
+```
+imports/
+  example-cluster.ts
```
eladb, line 177: Replace with usage example

packages/cdk8s-cli/README.md
```diff
@@ -32,26 +34,147 @@ $ cd my-fun-little-project
...
+
+```javascript
+import { cluster } from './imports/cluster';
+import { autoscaler } from './imports/autoscaler';
+```
+
+Python:
+
+```python
+from imports import cluster, autoscaler
```
campionfellin, line 121: This is the only thing I'm not entirely certain that holds true. I think we can write more python examples and fix this later if it's not accurate.

eladb: okay

PR #142, opened 2020-05-12
feat(cli): support CRDs with apiVersion "apiextensions.k8s.io/v1"

eladb:
Stop failing when a CRD uses the v1 api version. It is compatible with v1beta from our point of view and therefore can be imported.

Furthermore, ignore properties that are not camelCase since these can't be compiled to jsii. Users will have to use escape hatches for these.

Clean up some code and tests.


packages/cdk8s-cli/test/import/import-crd.test.ts
```diff
@@ -64,4 +51,17 @@ test('fails if one apiObject in multiObject CRD is not a valid CRD', async () =>
       kind: 'CustomResourceDefinition'
     }
   ])).toThrow('manifest does not have a "spec" attribute');
-})
\ No newline at end of file
+})
```
campionfellin, line 54:
```suggestion
});
```

packages/cdk8s-cli/lib/import/type-generator.ts
```diff
@@ -214,6 +214,11 @@ export class TypeGenerator {
         if (propName.startsWith('x-')) {
           continue; // skip extensions for now
         }
+
+        if (propName.includes('_')) {
```
campionfellin, line 218: Can you explain a bit more why this doesn't work with JSII? Are escape hatches supported in `cdk8s` yet?

packages/cdk8s-cli/lib/import/crd.ts
```diff
@@ -101,25 +112,26 @@ export class ImportCustomResourceDefinition extends ImportBase {
       return undefined;
     }
 
-    return yaml.parseAllDocuments(manifest)
-               .map((doc: yaml.ast.Document) => doc.toJSON())
-               .filter((doc) => doc && (doc as CustomResourceApiObject).kind === 'CustomResourceDefinition');
+    return yaml.parseAllDocuments(manifest).map((doc: yaml.ast.Document) => doc.toJSON());
```
campionfellin, line 115: I like this cleanup!

packages/cdk8s-cli/lib/import/type-generator.ts
eladb: jsii requires that names will be capitalized in a certain way in order to be able to case-convert them to different languages in an idiomatic way. We might be able to lift off that restriction, but will require a bit more investigation. @romainmuller any thoughts?

We don't have escape hatches yet (#144)

RomainMuller: Yep. I'd like to allow more freedom in naming but right now we rely on tight constraints to make things work. Lifting the restriction will actually take a significant effort so it will be a longer haul kind of thing.

Basically the way I think about it is that it's come as a benefit of moving name generation (for target languages) earlier in the value flow (compiler pre-names everything, instead of code generators doing this at the last minute).

PR #149, opened 2020-05-14
feat(cli): java support for "import"

campionfellin:
First pass here at Java support. It's fairly straightforward, but the PR is pretty bloated from the integ test files. I'll comment inline where relevant. Main concerns and uses:

#### Example project
```
language: java
app:  mvn exec:java -Dexec.mainClass="com.mycompany.app.HelloKube"
imports:
  - k8s
```

When importing, `k8s` looks like this:

```
imports/
  k8s/
    pom.xml
    src/main/java/k8s/ a bunch of .java files
    src/main/resources/k8s/k8s@0.0.0.jsii.tgz
```

Java code:

```java
package com.mycompany.app;

import org.cdk8s.Chart;
import software.constructs.Construct;
import k8s.Service;

public class HelloKube extends Chart 
{

    public HelloKube(final Construct scope, final String id, final ChartOptions options) {
        super(scope, id, options);

        // Defining a LoadBalancer Service
        final String serviceType = "LoadBalancer"; 
.
.
.
```

Concerns:

1. The integ tests seem to be almost no-ops due to my liberal usage of `--exclude` with the diffs. Similar to with Python, when JSII compiles the classes, it timestamps them. Thus the diff always fails.

2. When importing, the `pom.xml` for `k8s` has dependency version of `[undefined]` for both `software.constructs` and `org.cdk8s`

I'll try to make these comments inline below, so you can pick them apart from all the test files :)


packages/cdk8s-cli/bin/cmds/init.ts
```diff
@@ -67,13 +68,15 @@ async function determineDeps(version: string, dist?: string): Promise<Deps> {
     'npm_cdk8s': `cdk8s@${ver}`,
     'npm_cdk8s_cli': `cdk8s-cli@${ver}`,
     'pypi_cdk8s': `cdk8s~=${version}`, // no support for pre-release
+    'maven_cdk8s': `cdk8s -Dversion=${version}`,
```
campionfellin, line 71: I think this is how it should be, but I'm not too familiar with Maven.

packages/cdk8s-cli/lib/import/base.ts
```diff
@@ -74,13 +74,27 @@ export abstract class ImportBase {
...
       default:
         throw new Error(`unsupported language ${options.targetLanguage} (yet)`);
     }
   
   }
 
+  private async harvestJava(targetDir: string, moduleName: string) {
+    const target = path.join(targetDir, moduleName);
+    await fs.move(`dist/java/`, target, { overwrite: true });
+  }
```
campionfellin, line 96: Harvest java method

packages/cdk8s-cli/lib/import/jsii.ts
```diff
@@ -48,6 +48,13 @@ export async function jsiiCompile(workdir: string, options: JsiiCompileOptions)
     jsii: {
       outdir: "dist",
       targets: {
+        java: {
+          package: main,
+          maven: {
+            groupId: main,
+            artifactId: main
+          }
+        },
```
campionfellin, line 57: JSII config. Naming seemed reasonable. Produces:

```xml
  <groupId>k8s</groupId>
  <artifactId>k8s</artifactId>
  <version>0.0.0</version>
  <packaging>jar</packaging>
```

test/test-imports/java/expected-from-cli/clusterinstallation/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+    <dependency>
+      <groupId>org.cdk8s</groupId>
+      <artifactId>cdk8s</artifactId>
+      <version>[undefined]</version>
+    </dependency>
+    <dependency>
+      <groupId>software.constructs</groupId>
+      <artifactId>constructs</artifactId>
+      <version>[undefined]</version>
+    </dependency>
```
campionfellin, line 46: This is what I mean by `[undefined]` versions here

test/test-imports/python/cdk8s.yaml
```diff
@@ -1,5 +1,4 @@
-app: node index.js
-language: typescript
+language: python
```
campionfellin, line 1: Noticed this, just cleaning up

test/test-imports/java/test.sh
```diff
@@ -0,0 +1,28 @@
...
+cp ${scriptdir}/mattermost_crd.yaml .
+cp ${scriptdir}/example_multiple_crd.yaml .
+
+cdk8s import mattermost:=mattermost_crd.yaml --language java
+
+cdk8s import --language java
+
+# Exclude some files because their annotations differ
+# because they were pacmak-ed at different times.
+diff -r --exclude=*.java --exclude=*.xml imports ${scriptdir}/expected-from-config
```
campionfellin, line 18: Seems like I'm excluding everything here...

packages/cdk8s-cli/bin/cmds/init.ts
eladb: Isn't this part of the init template work? Revert here.

packages/cdk8s-cli/lib/import/jsii.ts
eladb: groupId and artifactId is fine. I am wondering about the java package. Feels odd that this will just go to "k8s". I think it should be `cdk8s.imports.k8s`, no?

eladb:
Update PR title.

packages/cdk8s-cli/bin/cmds/init.ts
campionfellin: Great point, I'll move this over to that PR

packages/cdk8s-cli/lib/import/jsii.ts
campionfellin: That seems reasonable to me! I'll switch it up.

campionfellin:
Opened up a guidance question with `jsii` regarding the `[undefined]`: https://github.com/aws/jsii/issues/1678

PR here: https://github.com/awslabs/cdk8s/pull/161 to more easily update snapshots

Will look into the integ tests next

campionfellin:
I think this is blocked by https://github.com/aws/jsii/issues/1678 for now. Do you think once I add integ tests back, we should push this just to give a start at some Java support?

eladb:
> I think this is blocked by https://github.com/aws/jsii/issues/1678 for now. Do you think once I add integ tests back, we should push this just to give a start at some Java support?

Why are we using a * dependency? We should use an explicit version.

campionfellin:
It comes from the dummy `package.json` that's setup here:

https://github.com/awslabs/cdk8s/blob/85cf6313c20771b718c19ee6085afc23cd787311/packages/cdk8s-cli/lib/import/jsii.ts#L46-L71

I can try to just pop in the latest version there

packages/cdk8s-cli/lib/import/base.ts
```diff
@@ -74,13 +74,27 @@ export abstract class ImportBase {
...
   
       default:
         throw new Error(`unsupported language ${options.targetLanguage} (yet)`);
     }
   
   }
 
+  private async harvestJava(targetDir: string, moduleName: string) {
+    const target = path.join(targetDir, moduleName);
+    await fs.move(`dist/java/`, target, { overwrite: true });
```
eladb, line 95: ok this is where I think I missed something: technically I think we should harvest only the sources (& jsii tarball) but not the pom file.

campionfellin: Aha! I'm new to maven, so I was having some troubles importing `k8s` like this:

```
    <dependency>
      <groupId>k8s</groupId>
      <artifactId>k8s</artifactId>
      <version>0.0.0</version>
    </dependency>
```

without the `pom.xml`...

Did some more digging, and I can do it without the `pom.xml` using:

```
    <resources>
      <resource>
        <directory>imports/k8s/main/resources</directory>
      </resource>
      <resource>
        <directory>imports/k8s/main/java</directory>
      </resource>
    </resources>
```

Changing above code to `await fs.move(`dist/java/src/`, target, { overwrite: true });` will work (and exclude the `pom.xml`) and in fact, brings this in line with what we're doing for python as well.

I'll refactor and combine the `harvestPython()` and `harvestJava()`

test/test-imports/java/test.sh
```diff
@@ -0,0 +1,39 @@
...
+cp ${scriptdir}/cdk8s.yaml  .
+cp ${scriptdir}/mattermost_crd.yaml .
+cp ${scriptdir}/example_multiple_crd.yaml .
+
+cdk8s import mattermost:=mattermost_crd.yaml --language java
+
+for f in `find imports/*`
+do
+    diff -I '^@javax.annotation.Generated' $f ${scriptdir}/expected-named-from-cli/${f#*/}
+done
```
campionfellin, line 17: @eladb I could not get `grep -v` working, however, this way of diffing worked for me. Let me know if it's alright!

campionfellin:
Alright, I think this is ready to go. @eladb let me know once you've reviewed and I'll push the test files :)

eladb:
@campionfellin Please hold off with this. I am working on extracting all the jsii logic into a separate library so we will move this code to there.

eladb:
@campionfellin sorry for all the mess...

Would you be interested in picking this up in https://github.com/eladb/jsii-srcmak?

eladb:
I am closing for now

campionfellin:
No problem! Yep, I'll pick it up and bring over to https://github.com/eladb/jsii-srcmak :)

eladb:
Thanks! And sorry again for shoving this over...

campionfellin:
No problem :)

PR #164, opened 2020-05-15
feat: initial support for OpenShift

mostafahussein:
This PR adds integration with OpenShift. I have the following points in order to decide how to continue.

1- There are some functions and variables in `k8s.ts` needs to be importable by `openshift.ts`, do you think that using a shareable file like `k8s-util.ts` will be better? I mean to avoid code duplication as possible between `openshift.ts` and `k8s.ts`. So any suggestions for my PR so I can enhance it are welcome
2- What would be the best approach to test this new import?


mostafahussein:
I guess all Kubernetes flavors are the same, the only difference will be the custom resources that each vendor adds to Kubernetes and the version that is being used as a base for their Kubernetes flavor. That's the case with OpenShift and with Rancher but I cannot find the rancher schema at the moment

eladb:
> all Kubernetes flavors are the same, the only difference will be the custom resources that each vendor adds to Kubernetes and the version that is being used as a base for their Kubernetes flavor. That's the case with OpenShift and with Rancher but I cannot find the rancher schema at the moment. So adding generic support will be better so it could be easier to import X and use it with cdk8s

Sounds good. So let's consider a more generic approach where you can point `cdk8s import` to a schema definition (such as [the one you used](https://github.com/garethr/openshift-json-schema/blob/master/master/_definitions.json)) and it will import from there.

I rather we didn't need to duplicate this for every k8s flavor. Does that make sense? I am closing this PR for now.

PR #168, opened 2020-05-17
feat(cli): jest tests in typescript-app template

mbonig:
Adding a basic jest test and jest configuration to typescript boilerplate.

---


packages/cdk8s-cli/templates/typescript-app/main.test.ts
```diff
@@ -0,0 +1,11 @@
+import {MyChart} from './main';
+import {App, Testing} from "cdk8s";
+
+describe('Placeholder', () => {
+    test('Empty', () => {
+        const app = new App();
```
eladb, line 6: Use `Testing.app()`

packages/cdk8s-cli/templates/typescript-app/main.test.ts
```diff
@@ -0,0 +1,11 @@
...
+        const chart = new MyChart(app, 'test-chart');
+        const results = Testing.synth(chart)
+        expect(results).toHaveLength(0);
```
eladb, line 9: I would show how to use snapshot tests:

```ts
expect(Testing.synth(chart)).toMatchSnapshot();
```

packages/cdk8s-cli/templates/typescript-app/main.test.ts
mbonig: changed.

packages/cdk8s-cli/templates/typescript-app/main.test.ts
mbonig: changed

mbonig:
Apparently there is something I'm really missing about the way the CI system builds and tests this. For example, right now it's failing because the of the snapshot test I added. From what I can tell, it shouldn't even be running the templated test. I'm going to mark this WIP until I understand this better.

packages/cdk8s-cli/templates/typescript-app/main.test.ts
mbonig: I updated this to have the .toMatchSnapshot() call but it's causing the build to fail. 

When the `init` happens it also builds and run tests. Those tests then fail because the `jest -u` wasn't passed, the snapshot won't be created unless specifically told to update (this doesn't appear to be a problem locally, though). Looking through the code it appears that after the project is scaff'd it'll do a 'build'. Before I go and rework the scaff so that it calls a special 'first time build' (with the -u), is there a better way to handle this IYO?

mbonig: Alright, I just replaced the 'npm run build' with a breakout of the individual commands and things are passing now.

mbonig:
Anything else here worth changing?

packages/cdk8s-cli/templates/typescript-app/jest.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    "roots": [
+        "<rootDir>"
+    ],
+    testMatch: [ '**/*.test.ts'],
```
eladb, line 5: is coverage enabled by default? Would be nice to include that.

PR #181, opened 2020-05-21
chore: fix lerna build globbing

campionfellin:
Recent changes to the `examples/` directory stopped building the example packages.

This fixes that.

Also described here: https://github.com/awslabs/cdk8s/pull/175

```
Lerna package globbing is overridden by yarn workspaces, so removed that and fixed globbing there. Docs - https://github.com/lerna/lerna/blob/master/commands/bootstrap/README.md#--use-workspaces
```

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,8 +1,4 @@
 {
-  "packages": [
-    "packages/*",
-    "examples/**/*"
-  ],
   "npmClient": "yarn",
   "useWorkspaces": true,
   "version": "0.0.0"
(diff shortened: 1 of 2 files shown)


lerna.json
```diff
@@ -1,8 +1,4 @@
 {
-  "packages": [
-    "packages/*",
```
eladb, line 3: I believe both are needed

campionfellin: I don't think so.

```
if --use-workspaces is true then packages will be overridden by the value from package.json/workspaces.
```

According to https://github.com/lerna/lerna/blob/master/commands/bootstrap/README.md#--use-workspaces

And we can see in the build it builds all 9 packages

eladb: Okay!

PR #183, opened 2020-05-22
fix(cli): import doesn't work on windows

campionfellin:
Might help with https://github.com/awslabs/cdk8s/issues/182 according to https://stackoverflow.com/questions/37459717/error-spawn-enoent-on-windows

(PS: currently working on adding `macos-latest` and `windows-latest` to github actions build).

### DIFF
--- a/packages/cdk8s-cli/lib/import/base.ts
+++ b/packages/cdk8s-cli/lib/import/base.ts
@@ -56,7 +56,7 @@ export abstract class ImportBase {
         });
 
         const pacmak = require.resolve('jsii-pacmak/bin/jsii-pacmak');
-        await shell(pacmak, [ '--target', options.targetLanguage, '--code-only' ]);
+        await shell(pacmak, [ '--target', options.targetLanguage, '--code-only' ], { shell: true });
         await this.harvestCode(options, outdir, name);
       });
     }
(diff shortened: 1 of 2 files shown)


packages/cdk8s-cli/lib/import/base.ts
```diff
@@ -56,7 +56,7 @@ export abstract class ImportBase {
         });
 
         const pacmak = require.resolve('jsii-pacmak/bin/jsii-pacmak');
-        await shell(pacmak, [ '--target', options.targetLanguage, '--code-only' ]);
```
eladb, line 59: Since both `jsii` and `pacmak` are node programs, I rather we changed `shell` to use `spawnFile`:

```
spawnFile(process.execPath, [ pacmak, '--target', options.targetLanguage, '--code-only' ]);
```

See https://github.com/awslabs/cdk8s/blob/0bf8dfd6a8346b8b2f0e33eac0a2ed32d384c387/packages/cdk8s/lib/yaml.ts#L74 as an example.

campionfellin: Works for me! I'll update it

eladb: Hey, I am realizing this PR is also going to be superseded by my extraction of jsii logic into a new library (you can check it out in https://github.com/eladb/jsii-srcmak)

eladb:
This is resolved by jsii-srcmak, so no need to fix anymore.

campionfellin:
Yay! Awesome :)

PR #198, opened 2020-05-24
feat(lib): yaml utility functions

eladb:
Introduce `Yaml.save(file, docs)`, `Yaml.tmp(docs)` and `Yaml.load(urlOrFile)` to standardize all YAML save/load operations.


packages/cdk8s/lib/yaml.ts
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Saves a set of YAML documents into a temp file (in /tmp)
+   *
+   * @returns the path to the temporary file
+   * @param docs the set of documents to save
+   */
+  public static tmp(docs: any[]): string {
+    const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), 'cdk8s-'));
+    const filePath = path.join(tmpdir, 'temp.yaml');
+    this.save(filePath, docs);
```
iliapolo, line 32: `this` inside a static method? does this translate well to Java/C#?

```suggestion
    Yaml.save(filePath, docs);
```

packages/cdk8s/lib/yaml.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    return result;
+  }
+
+  /**
+   * Utility class.
+   */
+  private constructor() {
+    return;
```
iliapolo, line 67: why `return` in a constructor?

packages/cdk8s/lib/api-object.ts
```diff
@@ -69,12 +69,13 @@ export interface ApiObjectOptions {
...
 export class ApiObject extends Construct {
-
   /**
-   * The app-unique name of the object.
+   * The name of the API object.
    *
-   * The name is allocated based on the path of the object construct within the
-   * construct tree.
+   * If a name is specified in `metadata.name` this will be the name returned.
+   * Otherwise, a name will generated by calling
```
iliapolo, line 76:
```suggestion
   * Otherwise, a name will be generated by calling
```

packages/cdk8s/lib/yaml.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+
+function readOrDownload(url: string) {
+  if (isFile(url)) {
+    return fs.readFileSync(url, 'utf-8');
+  } else {
+    return execSync(`curl --no-fail -L ${url}`, { encoding: 'utf-8' });
```
iliapolo, line 76: Any reason not to use `TypeScript` here instead of bashing out?

packages/cdk8s/lib/yaml.ts
iliapolo: Oh wait, this doesn't actually get **executed** in other languages. Still though...

eladb: Yes, totally works in JavaScript/TypeScript (this is the implementation, no it's not relevant for java/C#).

packages/cdk8s/lib/yaml.ts
eladb: node.js doesn't have a synchronous http request facility, so we need to use `execSync`. I might be able to `execSync` a node.js program but I figured `curl` is ubiquitous enough.

packages/cdk8s/lib/yaml.ts
eladb: Satisfy tslint who doesn't like empty methods.

packages/cdk8s/lib/yaml.ts
iliapolo: Forgot about the `sync` limitation in constructs. This will however break windows for sure. Also some debian distros only come with `wget` as far as I recall. I think its better to `execSync` a node program

eladb: ok, let me see what I can do ;-) thanks for keeping me honest

iliapolo: 👍

eladb: Ok, fixed.

PR #201, opened 2020-05-24
fix(lib): unable to express empty objects and array

eladb:
We used to remove all empty objects and array from synthesized objects. This is a problem in the k8s world since empty objects are sometimes used to represent a value.

For example, the [`volumes.emptyDir`](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) field.

- [ ] Wait for https://github.com/awslabs/cdk8s/pull/200

----


iliapolo:
@eladb I believe the body and title of this PR are not correct...? Its exactly the same as #200

PR #218, opened 2020-05-30
chore: bump min python version to 3.7.7

iliapolo:
Strictly speaking, any python `3.7.X` version should work, however, we've seen [issues](https://github.com/awslabs/cdk8s/issues/166) when using the **stock** python 3.7.3 version of MacOS. 

Those issues were resolved when upgrading to `3.7.7`, which is the latest *bugfix* release of the `3.7` minor version line. 

This PR bumps the minimum python version requirement in the getting started to `3.7.7`, as more of a safe-guard than anything else.  

### DIFF
--- a/docs/getting-started/python.md
+++ b/docs/getting-started/python.md
@@ -4,7 +4,7 @@ Let's walk through a simple "Hello, World!" example in Python.
 
 ## Prerequisites
 
-* [Python >= 3.7](https://www.python.org/downloads/release/python-377)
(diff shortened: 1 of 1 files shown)


iliapolo:
@eladb I'm ok with bumping this requirement. Python 3.7.3 is no longer available on brew (running `brew install python` will result in `3.7.7`), and it might very well be that `3.7.3` was somewhat buggy. Using 3.7.7 doesn't seem like too much of a hassle to users.

Let me know if you think differently

PR #221, opened 2020-06-02
chore: upgrade dependencies

campionfellin:

### DIFF
--- a/packages/cdk8s-cli/templates/python-app/Pipfile
+++ b/packages/cdk8s-cli/templates/python-app/Pipfile
@@ -5,3 +5,6 @@ verify_ssl = true
 
 [requires]
 python_version = "3.7"
+
+[packages]
+constructs = "==2.0.1"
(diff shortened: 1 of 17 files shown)


packages/cdk8s-cli/templates/python-app/Pipfile
```diff
@@ -5,3 +5,6 @@ verify_ssl = true
 
 [requires]
 python_version = "3.7"
+
+[packages]
+constructs = "==2.0.1"
```
campionfellin, line 10: @eladb had to pin this down, otherwise it was going for latest version. 

Generally, I don't think we want to worry about dependencies like this, but `constructs 3.0.1` introduced some breaking/incompatible changes and this seems to work for now.

campionfellin:
Opened a PR to fix mergify: https://github.com/awslabs/cdk8s/pull/224

PR #222, opened 2020-06-02
fix(docs): python interpreter line might need to be manually changed

trevorrobertsjr:
Users may encounter errors with running the tutorial as-is if the Python 3.7 is python3 rather than python in their operating system.


eladb:
Should we just change the interpreter lines to `python3` by default?

trevorrobertsjr:
Hello @eladb. I do not think so. Some users associate "python" with Python 3. So, changing the documentation to reference the python3 by default may cause issues?

PR #223, opened 2020-06-02
feat(lib): dependencies and ordering of charts and objects

### LINKED ISSUE #111: Dependencies and order between resources and charts
It is not uncommon to require that resources will be applied to a k8s cluster in a specific order. For example, if a resource is placed in a namespace, the `Namespace` resource must be applied before it.

Constructs have a concept of dependencies (`Node.of(this).addDependency(that)`), and cdk8s should respect dependencies between resources when the chart's manifest is synthesized so that the resources are ordered in the manifest based on their topological order.

Furthermore, if a dependency is added between two charts or resources from two different charts, we need to somehow codify the order *between the charts*. To do that we can employ a convention in which charts have a numeric ordinal prefix (i.e. `000-chartname.yaml`). This seems like a practice used by other projects ([example](https://github.com/IBM/composable/tree/master/releases/v0.1.4)) and offers a simple way for users to apply these charts through a simple `kubectl apply -f DIR`.

iliapolo:
`ApiObject` and `Chart` can now declare dependencies between one another. 

BREAKING CHANGE: `cdk8s` discontinues support for the `onPrepare` and `onSynthesis` construct hooks. These methods will eventually be removed from the `constructs` programming model.

Resolves #111

------------------------------------------------------------------------------------


iliapolo:
@eladb still have some touchups to do, but its ready for a provisional review. 

I'd especially like your feedback on: 

https://github.com/iliapolo/cdk8s/blob/3143769cbd90a9197ca2fe9f66f9b9cd0c7a11ec/packages/cdk8s/lib/dependency.ts#L42-L47

I went back and forth about cross-scope dependencies and eventually landed on limiting the graph to only include nodes that belong to the requested scope. Essentially describing the dependency relationships of the current construct sub-tree. It feels cleaner to me and perfectly fits into the chart dependency use-case, though it might be a bit counter intuitive.

packages/cdk8s/lib/app.ts
```diff
@@ -28,6 +32,39 @@ export class App extends Construct {
     this.outdir = options.outdir ?? process.env.CDK8S_OUTDIR ?? 'dist';
   }
 
+  protected onPrepare() {
```
eladb, line 35: Let's merge these to into `synth` and avoid using synthesis altogther?

packages/cdk8s/lib/app.ts
```diff
@@ -32,8 +32,13 @@ export class App extends Construct {
...
   }
 
-  protected onPrepare() {
-    super.onPrepare()
+  /**
+   * Synthesizes all manifests to the output directory
+   */
+  public synth(): void {
+    if (!fs.existsSync(this.outdir)) {
+      fs.mkdirSync(this.outdir);
```
eladb, line 40: Use `{ recurse: true }` instead of this `if`

packages/cdk8s/lib/app.ts
```diff
@@ -36,8 +40,32 @@ export class App extends Construct {
...
+      if (sourceChart !== targetChart) {
+        Node.of(sourceChart).addDependency(targetChart);
+      }
+    }
+
+    // TODO add validate call here
+
+    const charts: IConstruct[] = new DependencyGraph(Node.of(this)).topology().filter(x => x instanceof Chart);
+
+    for (const index in charts) {
```
eladb, line 59: It's a bit frowned upon to use `for-in`. I would just create a separate index:

```ts
let index = 0;
for (const chart of charts) {
  // do your magic.
  index++;
}
```

packages/cdk8s/lib/app.ts
```diff
@@ -36,8 +40,32 @@ export class App extends Construct {
...
+
+      const node: IConstruct = charts[index];
+      const chart: Chart = Chart.of(node);
+      const manifestFile = `${Node.of(chart).uniqueId}.k8s.yaml`;
+
+      const paddedIndex = index.toString().padStart(4, '0');
+
+      Yaml.save(path.join(this.outdir, `${paddedIndex}-${manifestFile}`), chart.toJson());
```
eladb, line 67: I think we shouldn't add the prefix if there is only a single chart or if there are no dependency relationship between charts.

packages/cdk8s/lib/app.ts
iliapolo: 👍

packages/cdk8s/lib/app.ts
iliapolo: 👍

packages/cdk8s/lib/app.ts
iliapolo: Done

packages/cdk8s/lib/app.ts
iliapolo: 👍

iliapolo:
@eladb This is ready for a second review round.

packages/cdk8s/README.md
```diff
@@ -76,6 +76,84 @@ NOTE: names of included objects (`metadata.name`) are preserved. This means that
...
+```console
+> cdk8s synth
+
+dist/0000-namespace.k8s.yaml
+dist/0001-application.k8s.yaml
+```
+
+Notice that the `namespace` chart appears first with the `0000` prefix. This will ensure that a subsequent execution of `kubectl apply -f dist/` will apply the `namespace` first, and the `application` second.
+
+#### Things just got cool
```
eladb, line 131: 🥰

packages/cdk8s/README.md
```diff
@@ -76,6 +76,84 @@ NOTE: names of included objects (`metadata.name`) are preserved. This means that
...
+
+```console
+> cdk8s synth
+
+dist/0000-namespace.k8s.yaml
+dist/0001-application.k8s.yaml
+```
+
+This means you need not be bothered with managing chart dependencies, simply work with the `ApiObject`s you create, and let `cdk8s` infer the chart dependencies.
+
```
eladb, line 156: We still don’t support deps between custom constructs correct? Worth mentioning and referencing an issue

packages/cdk8s/lib/app.ts
```diff
@@ -32,12 +36,61 @@ export class App extends Construct {
    * Synthesizes all manifests to the output directory
    */
   public synth(): void {
-    if (!fs.existsSync(this.outdir)) {
-      fs.mkdirSync(this.outdir);
+
+    fs.mkdirSync(this.outdir, { recursive: true });
+
+    const hasChartDependencies = this.inferChartDependencies();
```
eladb, line 42: Why can’t this be called inside produceManifest?

packages/cdk8s/lib/dependency.ts
```diff
@@ -0,0 +1,195 @@
...
+   * `null` in case this is the root of the graph.
+   */
+  public get value(): IConstruct | null {
+    return this._value;
+  }
+
+  /**
+   * Returns the children of the vertex (i.e dependencies)
+   */
+  public get children(): Set<DependencyVertex> {
```
eladb, line 112: Terminology: perhaps use “inbound” (dependents) and “outbound” (dependencies)

packages/cdk8s/test/app.test.ts
```diff
@@ -30,6 +31,58 @@ test('app with two charts', () => {
...
+  const chart2 = new Chart(app, 'chart2');
+  const chart3 = new Chart(app, 'chart3');
+
+  Node.of(chart1).addDependency(chart2);
+  Node.of(chart2).addDependency(chart3);
+
+  app.synth();
+
+  // THEN
+  expect(fs.readdirSync(app.outdir)).toEqual([
```
eladb, line 50: Add a test for when only part of the charts have dependencies between them

packages/cdk8s/test/chart.test.ts
```diff
@@ -93,6 +93,57 @@ test('synthesizeManifest() can be used to synthesize a specific chart', () => {
...
+    const chart1 = new Chart(app, 'chart1');
+
+    const obj1 = new ApiObject(chart1, 'obj1', { apiVersion: 'v1', kind: 'Kind1' });
+    const obj2 = new ApiObject(chart1, 'obj2', { apiVersion: 'v1', kind: 'Kind2' });
+    const obj3 = new ApiObject(chart1, 'obj3', { apiVersion: 'v1', kind: 'Kind3' });
+
+    Node.of(obj1).addDependency(obj2);
+    Node.of(obj2).addDependency(obj3);
+
+    expect(chart1.toJson()).toEqual([obj3.toJson(), obj2.toJson(), obj1.toJson()]);
```
eladb, line 110: Style: put each array element in a new line (especially here where it’s all about the order)

packages/cdk8s/README.md
iliapolo: Adding it

packages/cdk8s/lib/dependency.ts
iliapolo: Yeah lets go with that 👍

packages/cdk8s/test/app.test.ts
iliapolo: 👍

packages/cdk8s/test/chart.test.ts
iliapolo: 👍

packages/cdk8s/lib/app.ts
iliapolo: `produceManifests` is called after the validation part and I think the validation should be aware of the inferred chart dependencies. Originally this was just one big method, but I like this separation, even though I agree returning a `boolean` and then accepting it is a bit awkward.

packages/cdk8s/README.md
```diff
@@ -85,7 +217,6 @@ cdk8s bundles a set of test utilities under the `Testing` class:
   chart.
```
eladb, line 218: Can you also add documentation on how to use the new `Synthesis` methods?

packages/cdk8s/lib/app.ts
```diff
@@ -32,12 +37,140 @@ export class App extends Construct {
...
+   * Each element returned in the resulting array represents a different ApiObject
+   * that is contained in the chart.
+   *
+   * Note that the return value order is important. It is determined by the various dependencies between
+   * the constructs in the chart, where the first element is the one without dependencies, and so on...
+   *
+   * @returns An array of JSON objects.
+   * @param chart the chart to synthesize.
+   */
+  public static synthChart(chart: Chart): any[] {
```
eladb: Let's make this internal:

```suggestion
  /**
   * Synthesize a single chart.
   *
   * Each element returned in the resulting array represents a different ApiObject
   * that is contained in the chart.
   *
   * Note that the return value order is important. It is determined by the various dependencies between
   * the constructs in the chart, where the first element is the one without dependencies, and so on...
   *
   * @returns An array of JSON objects.
   * @param chart the chart to synthesize.
   * @internal
   */
  public static _synthChart(chart: Chart): any[] {
```

packages/cdk8s/lib/app.ts
```diff
@@ -32,12 +37,140 @@ export class App extends Construct {
...
    */
   public synth(): void {
-    if (!fs.existsSync(this.outdir)) {
-      fs.mkdirSync(this.outdir);
+
+    fs.mkdirSync(this.outdir, { recursive: true });
+
+    // this is kind of sucky, eventually I would like the DependencyGraph
+    // to be able to answer this question.
+    const hasChartDependencies = App.prepare(this);
```
eladb, line 45: Can you please move all these static methods as free-standing functions in a separate (internal) file called `synthesis.ts`

packages/cdk8s/README.md
```diff
@@ -76,6 +76,84 @@ NOTE: names of included objects (`metadata.name`) are preserved. This means that
 if you try to include the same manifest twice into the same chart, your manifest
 will have duplicate definitions of the same objects.
 
+### Dependencies
+
+You can declare dependencies between various `cdk8s` constructs by using the built-in support of the underlying `constructs` model.
+
+#### ApiObjects
+
+For example, you can force kubernetes to first apply a `Service` before applying the `Deployment` that uses that service:
```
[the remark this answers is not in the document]
iliapolo: Yeah good call out, I'll see if I can quickly find another example, if not i'll use namespace .

packages/cdk8s/lib/api-object.ts
```diff
@@ -109,6 +109,16 @@ export class ApiObject extends Construct {
     this.name = options.metadata?.name ?? this.chart.generateObjectName(this);
   }
 
+  /**
+   * Create a dependency between this ApiObject and other constructs.
+   * These can be other ApiObjects, Charts, or Custom.
```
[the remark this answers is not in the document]
iliapolo: 👍

packages/cdk8s/lib/app.ts
```diff
@@ -32,12 +37,140 @@ export class App extends Construct {
...
+   * @returns An array of JSON objects.
+   * @param chart the chart to synthesize.
+   */
+  public static synthChart(chart: Chart): any[] {
+
+    const app: App = App.of(chart);
+
+    // we must prepare the entire app before synthesizing the chart
+    // because the dependency inference happens on the app level.
+    App.prepare(app);
```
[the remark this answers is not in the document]
iliapolo: Fair enough 👍

packages/cdk8s/lib/app.ts
iliapolo: Agreed on leaving them as internal module functions in `app.ts`

packages/cdk8s/lib/app.ts
iliapolo: 👍

packages/cdk8s/README.md
iliapolo: Is this relevant? We haven't added anything to testing capabilities, `Testing.synthChart` still calls `chart.toJson()`...

packages/cdk8s/lib/api-object.ts
```diff
@@ -109,6 +109,16 @@ export class ApiObject extends Construct {
...
+   *
+   * @param dependencies the dependencies to add.
+   */
+  public addDependency(...dependencies: IConstruct[]) {
```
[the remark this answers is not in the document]
iliapolo: Done

packages/cdk8s/lib/app.ts
```diff
@@ -32,12 +37,118 @@ export class App extends Construct {
...
+    resolveDependencies(app);
+
+    // validate the app since we want to call onValidate of the relevant constructs.
+    // note this will also call onValidate on constructs from possibly different charts,
+    // but thats ok too since we no longer treat constructs as a self-contained synthesis unit.
+    validate(app);
+
+    return chartToKube(chart);
+  }
+
```
iliapolo, line 93: Left this here since I don't think it clutters to much, and we might actually want to switch it to public at some point?

iliapolo:
@eladb I'm still not a 100% comfortable with the internal clutter, but its probably good enough to move on. A fresh look in a couple of weeks will do us better.

packages/cdk8s/lib/app.ts
```diff
@@ -32,12 +37,118 @@ export class App extends Construct {
...
+  private static of(c: IConstruct): App {
+
```
iliapolo, line 95: Left this here since I don't think it clutters to much, and we might actually want to switch it to public at some point?

packages/cdk8s/README.md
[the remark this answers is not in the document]
iliapolo: Switched to the namespace example

PR #233, opened 2020-06-11
chore: java web-service example

campionfellin:


examples/java/web-service/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+  <build>
+    <sourceDirectory>src/main/java</sourceDirectory>
+    <testSourceDirectory>src/test/java</testSourceDirectory>
+    <resources>
+      <resource>
+        <directory>imports/src/main/k8s/main/java</directory>
+      </resource>
+      <resource>
+        <directory>imports/src/main/k8s/main/resources</directory>
+      </resource>
```
eladb, line 52: Something feels wrong here. Is there any way we can change how import works so we don't need to add this for every new import?

campionfellin: I can look into it... wasn't seeming to work with just `imports/` unfortunately. Might be a Maven thing too

eladb: Let's try to find a way that pom.xml file does not need updates when importing new APIs.

campionfellin: Simplified it so there's only one place you need to add for every new import. Seems like this is how maven expects it to be:

https://maven.apache.org/plugins/maven-resources-plugin/examples/resource-directory.html

eladb: Why can't we just add all the imports under the same source tree, so there is no need to update the pom file at all for new imports?

campionfellin: Ahhh, now I understand. I'll go ahead and do that!

examples/java/hello/pom.xml
```diff
@@ -45,10 +45,7 @@
     <testSourceDirectory>src/test/java</testSourceDirectory>
     <resources>
       <resource>
-        <directory>imports/src/main/k8s/main/java</directory>
-      </resource>
-      <resource>
-        <directory>imports/src/main/k8s/main/resources</directory>
+        <directory>imports/src/main/resources</directory>
```
eladb, line 48: oh yeah!

examples/java/web-service/src/main/java/com/mycompany/app/WebService.java
```diff
@@ -0,0 +1,103 @@
...
+import k8s.Service;
+import k8s.ServiceOptions;
+import k8s.ServicePort;
+import k8s.ServiceSpec;
+import k8s.DeploymentSpec;
+import k8s.PodSpec;
+import k8s.Container;
+import k8s.ContainerPort;
+import k8s.Deployment;
+import k8s.DeploymentOptions;
```
eladb, line 28: I expected these to be `import imports.k8s.Service`... Pom.xml file still feels too complex.

campionfellin: Let me see if I can cut some stuff out now with the changes to `jsii-srcmak`

examples/java/web-service/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+  </properties>
+  <build>
+    <sourceDirectory>src/main/java</sourceDirectory>
+    <testSourceDirectory>src/test/java</testSourceDirectory>
+    <resources>
+      <resource>
+        <directory>imports/src/main/resources</directory>
```
eladb, line 42: Still weird... why are the imports not under `src/main/java/imports`?

campionfellin: Hmm, we could probably add a maven step to move them there. They get output to `imports/src/main`. See the output structure here:

https://github.com/eladb/jsii-srcmak/pull/3#issue-434169273

campionfellin:
```
imports/
  src/
    main/
      java/
        k8s/
        mattermost/
      resources/
        k8s/
        mattermost/
```

campionfellin:
Hey, nice, seems like https://github.com/awslabs/cdk8s/pull/224 is working!

---

Going to do java `cdk8s init` next

PR #240, opened 2020-06-18
fix: move output of java imports into /src/main/java

campionfellin:
As discussed offline, we want java imports to fall under `/src/main/java`. This PR does 2 things:

1. Fixes the outdir for java, so new file structure looks like this:

```
├── src
|  └── main
|     ├── java
|     |  ├── com
|     |  |  └── mycompany
|     |  └── imports
|     |     └── k8s
|     └── resources
|        └── imports
|           └── k8s
```

and you import with `import imports.k8s`

2. Also simplifies the `pom.xml`, with no need for maven magic

### DIFF
--- a/examples/java/hello/pom.xml
+++ b/examples/java/hello/pom.xml
@@ -9,13 +9,6 @@
   <url>http://maven.apache.org</url>
   <dependencies>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>3.8.1</version>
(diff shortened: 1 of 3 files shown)


campionfellin:
By the way, this does still create an empty `imports/` directory at the base of the project. I'm not too concerned about it for now, but I can look into it a bit more

campionfellin:
Great how a 5 minute call saved like ~1 week of back-and-forths on PR 😎

PR #247, opened 2020-06-23
chore: bump jsii version

iliapolo:
Needed to nuget publishing due to: https://github.com/aws/jsii/commit/9ef8f17e11fa3403043c3237463c16e08020c696


packages/cdk8s-plus/.projenrc.js
```diff
@@ -48,4 +48,10 @@ project.addScripts({
   build: 'jsii --silence-warnings=reserved-word && yarn docgen'
 });
 
+// this has to be done so because otherwise its gets overridden
+// by the jsii-release version configured in projen.
+project.addDevDependencies({
+  'jsii-release': Semver.caret('0.1.8'),
```
eladb, line 54: jsii-release is only needed in the root package.json, why do we need it here as well?

iliapolo: yeah you're right. I'll change it.

iliapolo: @eladb Done

iliapolo:
Closing as this upgrade is currently not possible due to jsii version conflicts with constructs

PR #249, opened 2020-06-23
chore: pin constructs dep

iliapolo:
Using a caret dependency in a module built with `jsii < 1.3.2` results in an invalid version range in the package manifest file for dotnet artifacts:

See: `[2.0.1, 3.0.0-0)`

```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
  <metadata>
    <id>Org.Cdk8s</id>
    <version>0.25.0</version>
    <authors>Amazon Web Services</authors>
    <owners>Amazon Web Services</owners>
    <requireLicenseAcceptance>false</requireLicenseAcceptance>
    <license type="expression">Apache-2.0</license>
    <licenseUrl>https://licenses.nuget.org/Apache-2.0</licenseUrl>
    <description>Cloud Development Kit for Kubernetes (Stability: Experimental)</description>
    <tags>cdk kubernetes k8s constructs</tags>
    <repository type="git" url="https://github.com/awslabs/cdk8s.git" />
    <dependencies>
      <group targetFramework=".NETCoreApp3.1">
        <dependency id="Amazon.JSII.Runtime" version="[1.1.0, 2.0.0)" exclude="Build,Analyzers" />
        <dependency id="Constructs" version="[2.0.1, 3.0.0-0)" exclude="Build,Analyzers" />
      </group>
    </dependencies>
  </metadata>
</package>
```

This results in the following error when trying to publish:

```console
error: Response status code does not indicate success: 400 (The NuGet package contains an invalid .nuspec file. The error encountered was: 'The package manifest contains an invalid Version: '3.0.0-0''. Correct the error and try again.)
```

This issue was [fixed](https://github.com/aws/jsii/commit/9ef8f17e11fa3403043c3237463c16e08020c696) in `jsii` a while back. The problem is that we cannot upgrade the `jsii` version because it will conflict with the `jsii` version of `constructs`. We can't upgrade `constructs` because it introduces some breaking changes with regards to tokenization, and would require some non trivial effort.

As a work around, I suggest to *pin* constructs dependency to `2.0.1`, avoiding the whole caret invalid version range issue.
In practice, this pinning has no real affect since the next version of constructs is `3.0.0`, there is no version in between. 

### DIFF
--- a/examples/java/hello/package.json
+++ b/examples/java/hello/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "cdk8s": "0.0.0",
-    "constructs": "^2.0.1"
+    "constructs": "2.0.1"
   },
   "devDependencies": {
     "cdk8s-cli": "0.0.0"
--- a/examples/python/crd/package.json
+++ b/examples/python/crd/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "cdk8s": "0.0.0",
-    "constructs": "^2.0.1"
+    "constructs": "2.0.1"
   },
   "devDependencies": {
     "cdk8s-cli": "0.0.0"
--- a/examples/python/hello/package.json
+++ b/examples/python/hello/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "cdk8s": "0.0.0",
-    "constructs": "^2.0.1"
+    "constructs": "2.0.1"
   },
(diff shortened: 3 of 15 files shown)


iliapolo:
> Mmmkay

Hhh yeah, that is definitely the correct sentiment for this :)

PR #257, opened 2020-06-30
fix: java importing crd fails

campionfellin:
Fixes https://github.com/awslabs/cdk8s/issues/234

Replace `/` with `.`

Tested locally with the CRD example I'm writing for Java.

### DIFF
--- a/packages/cdk8s-cli/lib/import/base.ts
+++ b/packages/cdk8s-cli/lib/import/base.ts
@@ -44,7 +44,7 @@ export abstract class ImportBase {
 
     for (const name of this.moduleNames) {
       // output the name of the imported resource
-      console.error(name);
+      console.log(name);
 
       const fileName = moduleNamePrefix ? `${moduleNamePrefix}-${name}.ts` : `${name}.ts`;
       code.openFile(fileName);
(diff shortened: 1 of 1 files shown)


packages/cdk8s-cli/lib/import/base.ts
```diff
@@ -44,7 +44,7 @@ export abstract class ImportBase {
 
     for (const name of this.moduleNames) {
       // output the name of the imported resource
-      console.error(name);
+      console.log(name);
```
eladb, line 47: Is that intentional?

packages/cdk8s-cli/lib/import/base.ts
```diff
@@ -87,9 +87,10 @@ export abstract class ImportBase {
 
           // java!
           if (options.targetLanguage === Language.JAVA) {
+            const javaName = name.replace('/', '.')
```
eladb, line 90: I never remember if this replaces all occurrences or just the first. I always use a reflex `/\//g`

packages/cdk8s-cli/lib/import/base.ts
campionfellin: Yeah, I thought the `error` was unintentional. Any reason for `error` over `log` here?

packages/cdk8s-cli/lib/import/base.ts
campionfellin: Will do that!

packages/cdk8s-cli/lib/import/base.ts
eladb: It depends whether we want people to be able to capture stdout and do stuff with it. I actually agree that this specific output should go to stdout so I am okay with this change

campionfellin: Sounds good 👍

campionfellin:
Will write tests for the base import method tomorrow (or later this week) and update the PR.

eladb:
> Will write tests for the base import method tomorrow (or later this week) and update the PR.

Okay let’s merge this in the meantime and add a test in a subsequent PR

PR #258, opened 2020-06-30
feat(lib): omit duplicate components in generated names

campionfellin:
For better readability, now omit adjacent duplicate components when generating resources names. Something like `redrive-redrive-app-configmap-e2d51b95` becomes `redrive-app-configmap-e2d51b95`.

Resolves https://github.com/awslabs/cdk8s/issues/236

BREAKING CHANGE: resource names will now be rendered differently, omitting adjacent duplicate components.

### DIFF
--- a/packages/cdk8s/src/names.ts
+++ b/packages/cdk8s/src/names.ts
@@ -51,9 +51,9 @@ export class Names {
 
     components.push(calcHash(path, HASH_LEN));
 
-    components = omitDuplicates(components);
-
     return components
+      .reverse()
+      .filter(omitDuplicates)
       .join('/')
       .slice(0, maxLen)
       .split('/')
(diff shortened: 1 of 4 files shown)


packages/cdk8s/src/names.ts
```diff
@@ -53,6 +53,7 @@ export class Names {
 
     return components
       .reverse()
+      .filter(((value, index, components) => value !== components[index-1]))
```
eladb, line 56: Extract the predicate to a function `omitDuplicates` so it’s more readable

packages/cdk8s/src/names.ts
```diff
@@ -67,6 +68,10 @@ export class Names {
   }
 }
 
+function omitDuplicates(components: string[]) {
```
campionfellin, line 71: Actually, will put this back so it's inline, using:

```
function omitDuplicates(value: string, index: number, components: string[]) {
```

PR #259, opened 2020-07-07
chore: link libs to local (port from cdk)

andrestone:
`link-all.sh` ported from aws-cdk (by creating symlinks from `rootDir` to `outDir` locally, for each package)


link-all.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
eladb, line 1: please put under `tools` and add an npm script in the root `package.json` file to invoke. Also update the contribution guide.

andrestone:
Updated integ references due to recent `jsii` changes (https://github.com/aws/jsii/pull/1738).

link-all.sh
andrestone: Maybe we shouldn't add it to package.json? This script is supposed to be called from other packages.

PR #270, opened 2020-07-15
WIP: escape hatches proposal

campionfellin:
## WIP

I based a large part of this proposal off of https://docs.aws.amazon.com/cdk/latest/guide/cfn_layer.html

I also remember previously (maybe I'm misremembering) that you could do escape hatches in `cdk` with raw YAML, which is why I included it here.

If it's easier to read this rendered out in markdown, I can put it on the readme for my fork as well.

Resolves https://github.com/awslabs/cdk8s/issues/144

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,52 +93,33 @@ const deployment = new kplus.Deployment(chart, 'Deployment', {
 });
 ```
 
-`cdk8s+` lets you set set replicas and the pod spec, but does not have an option for `strategy`. It leaves it to the default `RollingUpdate`. If you want to specify this, but still want to use `cdk8s+` constructs, you can use an override like so:
+At the moment, `cdk8s+` lets you define replicas and the pod spec, but does not have an option for `strategy`. It leaves it to the default `RollingUpdate`. If you want to specify this, but still want to use `cdk8s+` constructs, you can use an override like so:
 
 ```ts
-deployment.addOverride('spec.strategy', 'Recreate')
+import * as k8s from 'k8s'
+
+// Get the L1 construct behind the `cdk8s+` Deployment
+const childDeployment = deployment.node.defaultChild as k8s.Deployment
+
+// Set the DeploymentStrategy type
+childDeployment.spec.strategy.type = 'Recreate'
+
+// Alternatively, using a JSON path
+childDeployment.addOverride('spec.strategy.type', 'Recreate')
 ```
 
-#### Raw k8s Template Insertion
+In this case, your resulting spec might look like this:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,6 +66,80 @@ Choose your weapon:
...
+  spec: {
+    replicas: 3,
+    podSpecTemplate: {
+      containers: [ container ]
+    }
+  },
+});
+```
+
+`cdk8s+` lets you set set replicas and the pod spec, but does not have an option for `strategy`. It leaves it to the default `RollingUpdate`. If you want to specify this, but still want to use `cdk8s+` constructs, you can use an override like so:
```
eladb, line 96:
```suggestion
At the moment, `cdk8s+` lets you define replicas and the pod spec, but does not have an option for `strategy`. It leaves it to the default `RollingUpdate`. If you want to specify this, but still want to use `cdk8s+` constructs, you can use an override like so:
```

README.md
```diff
@@ -66,6 +66,80 @@ Choose your weapon:
...
+
+```ts
+import { Construct } from 'constructs';
+import { Chart, Raw } from 'cdk8s';
+
+class MyChart extends Chart {
+ constructor(scope: Construct, name: string) {
+    super(scope, name);
+
+    new Raw(this, 'raw', {
```
eladb, line 114: What's the difference between `Raw` and `Include`? I think they should be the same thing, no?

README.md
```diff
@@ -66,6 +66,80 @@ Choose your weapon:
...
+      containers: [ container ]
+    }
+  },
+});
+```
+
+`cdk8s+` lets you set set replicas and the pod spec, but does not have an option for `strategy`. It leaves it to the default `RollingUpdate`. If you want to specify this, but still want to use `cdk8s+` constructs, you can use an override like so:
+
+```ts
+deployment.addOverride('spec.strategy', 'Recreate')
```
eladb, line 99: Mention the format of the first argument (is is a JSON query?)

eladb, line 99: Similarly to the AWS CDK, I think overrides should be implemented in the L1 layer, not L2.

I am not sure we need explicit overrides in L2 (cdk8s+). Users should be able to use `node.defaultChild` to find the L1 behind an L2 and then we should add overrides to the L1 layer.

README.md
campionfellin: No difference, good callout! I actually didn't realize we had `Include` already :)

README.md
campionfellin: Ah, will do that. Yes, intended to be a json path

campionfellin: Aha gotcha, I'll take another look here then

PR #293, opened 2020-08-19
chore: Update java and typescript init templates to include cdk8s-plus dependency

campionfellin:
Ideally, all starter projects should use `cdk8s-plus`, so adding it as a default dependency here


campionfellin:
Weird...

```
ERROR: Could not find a version that matches jsii<2.0.0,>=1.7.0,~=1.1.0 (from constructs==2.0.2

There are incompatible versions in the resolved dependencies:
  jsii<2.0.0,>=1.7.0 (from constructs==2.0.2
```

packages/cdk8s-cli/templates/typescript-app/.hooks.sscaff.js
```diff
@@ -5,12 +5,13 @@ const constructs_version = require('../../package.json').dependencies.constructs
...
 exports.post = ctx => {
   const npm_cdk8s = ctx.npm_cdk8s;
+  const npm_cdk8s_plus = ctx.npm_cdk8s_plus;
   const npm_cdk8s_cli = ctx.npm_cdk8s_cli;
 
   if (!npm_cdk8s) { throw new Error(`missing context "npm_cdk8s"`); }
   if (!npm_cdk8s_cli) { throw new Error(`missing context "npm_cdk8s_cli"`); }
 
-  installDeps([ npm_cdk8s, `constructs@${constructs_version}` ]);
+  installDeps([ npm_cdk8s, npm_cdk8s_plus, `constructs@${constructs_version}` ]);
```
iliapolo, line 14: I think we need to add the same thing here: https://github.com/awslabs/cdk8s/blob/master/packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js#L24

iliapolo (changes requested):

packages/cdk8s-cli/templates/typescript-app/.hooks.sscaff.js
campionfellin: I've been looking at python here for a bit and can't figure it out. Maybe you have some ideas on this? Basically... in the integration tests, we point to the local distribution of our packages, so:

```
/Users/campionfellin/Desktop/quickarticle/cdk8s/dist/python
├── cdk8s-0.0.0-py3-none-any.whl
├── cdk8s-0.0.0.tar.gz
├── cdk8s-plus-0.0.0.tar.gz
└── cdk8s_plus-0.0.0-py3-none-any.whl
```

When it tries to install `cdk8s_plus-0.0.0-py3-none-any.whl` it finds a dependency on version `0.0.0` of `cdk8s`. It then goes to PyPi looking for `cdk8s` 0.0.0 (instead of locally) and we get this error:

```
Could not find a version that satisfies the requirement cdk8s<0.0.1,>=0.0.0 (from cdk8s-plus==0.0.0

(from versions: 0.11.0, 0.12.0, 0.13.0, 0.13.1, 0.14.0, 0.15.0, 0.15.1, 0.16.0, 0.17.0, 0.18.0, 0.19.0, 0.20.0, 0.21.0, 0.22.0, 0.23.0, 0.24.0, 0.25.0, 0.26.0, 0.27.0)
```

Any idea how to get Pipenv to ignore the transitive dependency or to point it elsewhere?

campionfellin: We point to the local `dist` here:

```
    const ret = {
      npm_cdk8s: path.resolve(dist, 'js', `cdk8s@${version}.jsii.tgz`),
      npm_cdk8s_cli: path.resolve(dist, 'js', `cdk8s-cli-v${version}.tgz`), // yarn pack adds a "v" before the version
      pypi_cdk8s: path.resolve(dist, 'python', `cdk8s-${version.replace(/-/g, '_')}-py3-none-any.whl`),
      mvn_cdk8s: path.resolve(dist, 'java', `org/cdk8s/cdk8s/${version}/cdk8s-${version}.jar`),
    };
```

https://github.com/awslabs/cdk8s/blob/master/packages/cdk8s-cli/src/cli/cmds/init.ts#L47

Which eventually gets piped through to here: https://github.com/awslabs/cdk8s/blob/master/packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js#L24

where we call:

```
pipenv install ./dist/python/cdk8s-0.0.0-py3-none-any.whl
```

We do something similar for Java:

https://github.com/awslabs/cdk8s/blob/master/packages/cdk8s-cli/templates/java-app/.hooks.sscaff.js#L25

Though I'm now realizing I need to update the Java scaffolding here as well.

I'll see what I can do here!

campionfellin: Took me a while here... but got it! Longer comment in the code explains what was going on.

examples/java/hello/pom.xml
```diff
@@ -11,7 +11,12 @@
     <dependency>
       <groupId>org.cdk8s</groupId>
       <artifactId>cdk8s</artifactId>
-      <version>0.25.0</version>
+      <version>0.26.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.cdk8s</groupId>
+      <artifactId>cdk8s-plus</artifactId>
```
iliapolo, line 18: I don't think we should add `cdk8s-plus` to an example that doesn't actually need it. Its good to have it as part of the template as an easy kick-off, but working examples can be left as is.

examples/java/web-service/pom.xml
```diff
@@ -11,7 +11,12 @@
     <dependency>
       <groupId>org.cdk8s</groupId>
       <artifactId>cdk8s</artifactId>
-      <version>0.25.0</version>
+      <version>0.26.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.cdk8s</groupId>
+      <artifactId>cdk8s-plus</artifactId>
```
iliapolo, line 18: Same

packages/cdk8s-cli/templates/java-app/.hooks.sscaff.js
```diff
@@ -14,7 +14,7 @@ exports.pre = (variables) => {
 };
 
 exports.post = options => {
-  const { mvn_cdk8s, cdk8s_version } = options;
+  const { mvn_cdk8s, mvn_cdk8s_plus, cdk8s_version } = options;
   if (!mvn_cdk8s) {
```
iliapolo, line 18: add `mvn_cdk8s_plus` to assertion

packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js
```diff
@@ -14,14 +14,35 @@ exports.pre = () => {
...
+  if (!pypi_cdk8s_plus) {
+    throw new Error(`missing context "pypi_cdk8s_plus"`);
+  }
 
   execSync('pipenv lock --clear')
-  execSync('pipenv install', { stdio: 'inherit' });
+  execSync('pipenv install --skip-lock', { stdio: 'inherit' });
   execSync(`pipenv install ${pypi_cdk8s}`, { stdio: 'inherit' });
+  /**
+   * Using --no-deps flag here to ignore subdependencies. For cdk8s_plus, that's
```
iliapolo, line 29: Was there a continuation for this sentence? :)

packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js
```diff
@@ -14,14 +14,35 @@ exports.pre = () => {
...
+   * Even when install locally, with this command:
+   * 
+   * pipenv install dist/python/cdk8s_plus-0.0.0-py3-none-any.whl
+   * 
+   * this would point to these sub-dependencies in https://pypi.org/ which
+   * unfortunataly does not have cdk8s-0.0.0. We've already installed that
+   * locally, so this is safe to ignore.
+   * 
+   */
+  execSync(`pip3 install --no-deps ${pypi_cdk8s_plus}`, { stdio: 'inherit' });
```
iliapolo, line 45: Add a check that `pip3` exists, similar to how we do for `pyenv`

iliapolo (changes requested):

packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js
campionfellin: Done!

packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js
campionfellin: Was meaning to list the dependencies. Updated the wording!

packages/cdk8s-cli/templates/java-app/.hooks.sscaff.js
campionfellin: Done

examples/java/web-service/pom.xml
campionfellin: Done

examples/java/hello/pom.xml
campionfellin: Done

iliapolo (approved):

Issue #323, opened 2020-09-22
`Node.of(construct).uniqueId` is not strictly compatible with the K8s labels


foriequal0:
### Description of the feature or enhancement:

I can find some usages of `Node.of(construct).uniqueId` across the codebase and examples. However, it is bound to CloudFormation's resource name rules and it isn't strictly compatable with Kubernetes' name/labels rules.
It might be compatible if the resource name is short enough to be within k8s' length limitations, but you still bear the fact that `Node.of(construct).uniqueId` gives you a `uniqueidbutwithbarelyreadablebecauseitlacksofseparatorcharactorsA8322FCE` when you use a snake-case for resources names.

There is already `Names.toDnsLabel` and `Chart.generateObjectName` using it. However, labels are superset of dns labels.

### Use Case:

I can find `uniqueId` in 8 files:

#### Potentially invalid
https://github.com/awslabs/cdk8s/blob/d863232e4bf8a6dfcdd11e32b18b03586a4b766d/packages/cdk8s-plus/src/deployment.ts#L208
They are examples, but it should be changed considering the fact that people are blindfoldingly conpy-paste them.
https://github.com/awslabs/cdk8s/blob/85cf6313c20771b718c19ee6085afc23cd787311/examples/typescript/podinfo/lib/deployment.ts#L92-L95
https://github.com/awslabs/cdk8s/blob/85cf6313c20771b718c19ee6085afc23cd787311/examples/typescript/web-service/web-service.ts#L33
https://github.com/awslabs/cdk8s/blob/0f70a4fa0a471b27b3c75e25d0609935060b44ed/docs/getting-started/typescript.md#L285

#### It would be better to be changed for the readibility
https://github.com/awslabs/cdk8s/blob/2ae890a92cf67be1e1a5478102dd6003e446440f/packages/cdk8s/src/app.ts#L51
https://github.com/awslabs/cdk8s/blob/2ae890a92cf67be1e1a5478102dd6003e446440f/packages/cdk8s/src/dependency.ts#L160
Okay-ish for internal use
https://github.com/awslabs/cdk8s/blob/2ae890a92cf67be1e1a5478102dd6003e446440f/packages/cdk8s/src/dependency.ts#L26-L32

### Proposed Solution:

It would be better to have a separate function for the label.

```typescript
...
const MAX_LABEL_LEN = 63;
const INVALIDATE_LABEL = /[^0-9a-zA-Z-_.]/
...
export class Names {
...
[...]

PR #326, opened 2020-09-23
fix(lib): `uniqueId` is not compatible with the k8s labels

foriequal0:
`Node.of(construct).uniqueId` is bound to CloudFormation's resource logical ID rule, which is not strictly compatible with the K8s label rule. It might lead to invalid K8s manifests.
I've changed all `Node.of(construct).uniqueId` with `Node.of(construct).path`, or `Names.toDnsLabel(Node.of(construct).path)` if needed.
Since K8s DNS label rule are subset of label key name segment/value rule, I've added `Names.toLabelValue` and used it for selector labels.

fixes: #323

BREAKING CHANGE: cdk8s-plus's value of a label `cdk8s.deployment` of Pods are changed

### DIFF
--- a/examples/typescript/cdk8s-plus-elasticsearch-query/__snapshots__/main.test.ts.snap
+++ b/examples/typescript/cdk8s-plus-elasticsearch-query/__snapshots__/main.test.ts.snap
@@ -128,13 +128,13 @@ function doSearch(uri, callback) {
       "replicas": 1,
       "selector": Object {
         "matchLabels": Object {
-          "cdk8s.deployment": "testchartDeploymentAAAF687B",
+          "cdk8s.deployment": "test-chart-deployment-98e15317",
         },
       },
       "template": Object {
         "metadata": Object {
           "labels": Object {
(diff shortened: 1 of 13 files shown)


foriequal0:
Rebase to resolve merge conflict.

examples/typescript/cdk8s-plus-elasticsearch-query/__snapshots__/main.test.ts.snap
```diff
@@ -128,13 +128,13 @@ function doSearch(uri, callback) {
       "replicas": 1,
       "selector": Object {
         "matchLabels": Object {
-          "cdk8s.deployment": "testchartDeploymentAAAF687B",
+          "cdk8s.deployment": "test-chart-Deployment-98e15317",
```
eladb, line 131: Should this be all-lowercase?

eladb: Strike that.

eladb (dismissed):

eladb (approved):

PR #327, opened 2020-09-24
fix: Fix yaml quote serialization 325

### LINKED ISSUE #325: Yaml files generated with synth do not quote strings which can be interpreted as native non-string types
### Description of the bug:

When `app.synth()` generates `yaml` files, values which are strings in memory are rendered without quotes in the `yaml` files. This can cause problems for string values such as `"true"`, `"yes"`, and `"2020-01-01"`. These strings, when rendered without quotes, will be read as native data types by `yaml` parsers such as `PyYaml` (`true` and `yes` become python `bool`, while `2020-01-01` becomes `datetime.datetime`).

### Reproduction Steps:

```py
import cdk8s
import yaml

# Create the chart with the unquoted values
app = cdk8s.App()
chart = cdk8s.Chart(app, "test")
cdk8s.ApiObject(chart, "foo", api_version="v1", kind="Foo", metadata={"annotations": {"bar": "2020-01-01", "baz": "on"}})
app.synth()

# Reread the generated file with pyyaml
with open('dist/test.k8s.yaml', 'r') as f:
    objects = list(yaml.safe_load_all(f))

# Print out some info about the re-read types
for key in ["bar", "baz"]:
    print(f"--> {key}")
    print(objects[0]["metadata"]["annotations"][key])
    print(type(objects[0]["metadata"]["annotations"][key]))
```

**Output**
```
--> bar
2020-01-01
<class 'datetime.date'>
--> baz
True
<class 'bool'>
```

**Rendered yaml**
```yaml
apiVersion: v1
kind: Foo
metadata:
  annotations:
    bar: 2020-01-01
    baz: on
  name: test-foo-657aa33b
```

### Error Log:

There are no immediate errors. See the rendered yaml and script output above.

### Environment:

  - **Framework Version**:

```
pip freeze | grep cdk8s
cdk8s==0.27.0
```

  - **OS**: Mac OS, `conda` w/ python 3.6

### Other:

--- 

This is :bug: Bug Report

gabe-l-hart:

## Description

This PR addresses https://github.com/awslabs/cdk8s/issues/325 by upgrading the version of the `yaml` package and enabling the option to use double-quotes when serializing strings.

Fixes #325 

## Testing

A unit test has been added and existing tests have been updated to ensure that string values are quoted correctly.

## Known Problems

Upgrading the version of `yaml` has broken the current functionality for serializing `undefined` values. The current test for `undefined` expects that it will be converted to `null`. Version `2.0.0-0` of `yaml` instead drops these values. [On the current `master`](https://github.com/eemeli/yaml/tree/master), there are two ways that this could be fixed: (1) set the [`keepUndefined` option](https://github.com/eemeli/yaml/blob/master/docs/03_options.md#yamldocumentdefaults) or (2) implement [a custom `replacer`](https://github.com/eemeli/yaml/blob/master/docs/02_parse_stringify.md#yamlstringifyvalue-replacer-options---string). 

Neither of these methods is available in `2.0.0-0`, so this PR will likely not be usable until the next release of `yaml` is out with those features enabled.


packages/cdk8s/package.json
```diff
@@ -61,7 +61,7 @@
   "dependencies": {
     "follow-redirects": "^1.11.0",
     "json-stable-stringify": "^1.0.1",
-    "yaml": "^1.7.2"
+    "yaml": "^2.0.0-0"
```
eemeli, line 64:
```suggestion
    "yaml": "2.0.0-0"
```
I would strongly recommend against using a `^` range with prerelease versions. `2.0.0-0` should be usable even in production, but I make no guarantees that the API won't change in `2.0.0-1`.

eemeli (changes requested):

packages/cdk8s/package.json
eladb: I am okay with that

packages/cdk8s/src/yaml.ts
```diff
@@ -15,6 +20,12 @@ export class Yaml {
    */
   public static save(filePath: string, docs: any[]) {
     // convert each resource to yaml and separate with a '---' line
+    //
+    // NOTE: the `keepUndefined` flag is not supported in the release of yaml
+    //  but is present on the tip of the `master` branch, so this does not work
+    //  currently, but it will when the next release of the yaml package is
+    //  out.
+    // const data = docs.map(r => YAML.stringify(r, {keepUndefined: true})).join('---\n');
```
eladb, line 28: So what's the plan?

packages/cdk8s/package.json
gabe-l-hart: Yes, good point. Thanks for the suggestion

packages/cdk8s/src/yaml.ts
gabe-l-hart: I think for now, we need to hold on this fix until the `yaml` version has been released to include `keepUndefined`. @eemeli It looks like you're a/the primary maintainer for the `yaml` package. Do you have a timeline for when a version will be released with `keepUndefined`?

eladb: Maybe we can convert all `undefined`s to `null`s?

eemeli: Soon-ish, esp. as there's clearly a need for it. :) I'm also working on some pretty intensive parser changes, so it'd make sense to wrap up the current state of affairs into a release before those get incorporated. For more specificity, maybe within a week or so?

gabe-l-hart: @eemeil, I think within a week should work for my team's timing. Thanks!

@eladb I'll update this PR once the `yaml` version is published. I think that's probably better than working around the issue in the short term since this feature has already been implemented in the library.

gabe-l-hart:
@eladb I do see that the [`integration tests` section of the build](https://github.com/awslabs/cdk8s/pull/327/checks?check_run_id=1188349677) is failing. It looks to me like the diff it shows has to do with a failing snapshot test in the `imports` functionality. I don't _think_ the change in this PR should have effected that section of code, but I could imagine that rebuilding the `yarn.lock` could have accidentally bumped the version of a different package and caused this failure. Do you have any more insight on what would have caused this test to fail? Thanks!

gabe-l-hart:
@eladb I was able to figure out the failures. Syncing up with the latest `master` fixed integration tests. I did need to update the integration snapshots to include quotes around the serialized yaml file strings which is expected with this change.

eladb (approved):

packages/cdk8s/src/yaml.ts
eemeli: [`yaml@2.0.0-1`](https://github.com/eemeli/yaml/releases/tag/v2.0.0-1) is now available.

eladb: Cool, @gabe-l-hart let me know when ready

gabe-l-hart: @eemeli Thank you! I'll make the updates and make sure the tests look good

gabe-l-hart:
@eladb I've updated for the latest `yaml` version, enabled the `keepUndefined` flag (with handling for fully `undefined` docs), and updated the tests to match their original checks. I think this will need another ✅  check from you before being ready to merge. Thanks!

eladb (approved):

PR #329, opened 2020-09-24
feat(lib): Expose DependecyGraph for upstream use

### LINKED ISSUE #328: Custom `synth` - Access to underlying Dependency Graph

cdk8s should expose programatic access to the underlying dependency graph so that it can be walked with code without having to read and write yaml from disk. 

### Use Case:

I'm implementing a python-based operator framework based on CDK8s. As such I would like to, given an App, walk the dependency graph applying api objects in the correct order based on dependencies. 

I can currently do this by `synth`ing to a temporary dir, reading the results and applying from there, however it adds an extra error prone (#325) step and prevents me from knowing inter-chart dependencies. 

### Proposed Solution:

Expose the `DependencyGraph` as a public export. This is clearly and advanced use-case, I'm happy with advanced APIs, but I'd prefer to not have to manually translate the TS into python 😛. 

---

barlock:

Fixes #328 

Just exposing this so I can use it in Python


barlock:
I suppose https://github.com/aws/jsii/issues/1 is blocking this one. 🤔 Need to figure out if an Array will do.

barlock:
Settled on using sets for underlying implementation, but exposing Arrays. Best of both worlds I think.

eladb (commented):
LGTM, @iliapolo what do you think?

iliapolo (dismissed):
@barlock Looks good. Notice the build fails on the integration tests now.

Let us know if you need help tackling them.

barlock:
Took me a while to get `dotnet` setup in order run the packaging steps. 

Apologies for what seems like it's turned into a massive pr? The snapshots seem to be updating a stylistic thing? If you can let me know what's safe to revert from those snapshots I'll happily do it to reduce the size.

iliapolo (approved):
@barlock No worries - looks like we don't lock some `jsii` packages in our integration tests.

Thanks for pushing this through!

PR #342, opened 2020-10-14
feat(plus): service.addDeployment()

eladb:
Introduce `service.addDeployment()` which associates a deployment to a service. The method `deployment.expose()` now just delegates the work to this method.

Additionally, the logic that added the selector label to the deployment was moved from `deployment._toKube()` into `service.addDeployment()` and combined with the logic that installed the selector on the service (which is much more natural).

BREAKING CHANGE: `deployment.expose()` now takes `port` as a positional argument (before: `deployment.expose({ port })`, now: `deployment.expose(port)`).

### DIFF
--- a/packages/cdk8s-plus/API.md
+++ b/packages/cdk8s-plus/API.md
@@ -326,6 +326,7 @@ new Deployment(scope: Construct, id: string, props?: DeploymentProps)
 * **id** (<code>string</code>)  *No description*
 * **props** (<code>[DeploymentProps](#cdk8s-plus-deploymentprops)</code>)  *No description*
   * **metadata** (<code>[ApiObjectMetadata](#cdk8s-apiobjectmetadata)</code>)  Metadata that all persisted resources must have, which includes all objects users must create. __*Optional*__
+  * **defaultSelector** (<code>boolean</code>)  Automatically allocates a pod selector for this deployment. __*Default*__: true
   * **spec** (<code>[DeploymentSpec](#cdk8s-plus-deploymentspec)</code>)  The spec of the deployment. __*Default*__: An empty spec will be created.
 
 
@@ -341,18 +342,18 @@ Name | Type | Description
 ### Methods
 
 
-#### expose(options)🔹 <a id="cdk8s-plus-deployment-expose"></a>
+#### expose(port, options?)🔹 <a id="cdk8s-plus-deployment-expose"></a>
 
 Expose a deployment via a service.
 
 This is equivalent to running `kubectl expose deployment <deployment-name>`.
 
 ```ts
-expose(options: ExposeOptions): Service
+expose(port: number, options?: ExposeOptions): Service
 ```
 
-* **options** (<code>[ExposeOptions](#cdk8s-plus-exposeoptions)</code>)  - Options.
-  * **port** (<code>number</code>)  The port number the service will bind to. 
+* **port** (<code>number</code>)  The port number the service will bind to.
+* **options** (<code>[ExposeOptions](#cdk8s-plus-exposeoptions)</code>)  Options.
   * **serviceType** (<code>[ServiceType](#cdk8s-plus-servicetype)</code>)  The type of the exposed service. __*Default*__: ClusterIP.
 
 __Returns__:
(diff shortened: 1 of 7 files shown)


packages/cdk8s-plus/API.md
```diff
@@ -805,6 +805,21 @@ addRules(...rules: IngressRule[]): void
 
 
 
+#### protected onValidate()🔹 <a id="cdk8s-plus-ingress-onvalidate"></a>
```
iliapolo, line 808: Where did this come from suddenly?

packages/cdk8s-plus/src/deployment.ts
```diff
@@ -195,21 +176,11 @@ export class DeploymentSpecDefinition {
...
     return { ...this._labelSelector };
   }
-
+  
   /**
    * @internal
    */
-  public _toKube(deployment: Deployment): k8s.DeploymentSpec {
-
-    // automatically select pods in this deployment
```
iliapolo, line 204: So how are the pods in the deployment selected now?

packages/cdk8s-plus/src/service.ts
```diff
@@ -272,6 +272,37 @@ export class ServiceSpecDefinition {
...
+    const containers = deployment.spec.podSpecTemplate.containers;
+    if (containers.length === 0) {
+      throw new Error('Cannot expose a deployment without containers');
+    }
+
+    // create a label and attach it to the deployment pods
+    const selector = 'cdk8s.deployment';
+    const matcher = Names.toLabelValue(Node.of(deployment).path);
+
+    deployment.spec.podMetadataTemplate.addLabel(selector, matcher);
```
iliapolo, line 295: I dont think this belongs here - why are pods only selected if the deployment is added to a service?

packages/cdk8s-plus/src/service.ts
```diff
@@ -272,6 +272,37 @@ export class ServiceSpecDefinition {
...
+   * Exposes a deployment through this service.
+   *
+   * Requests will be routed to the port exposed by the first container in the
+   * deployment's pods. The label `cdk8s.deployment` will be added to pods in
+   * the deployment and used to select pods by this service.
+   *
+   * @param deployment The deployment to expose
+   * @param port The external port
+   */
+  public addDeployment(deployment: Deployment, port: number) {
```
iliapolo, line 285: I'm curious why you decided to implement it like this. Also - is there meaning to adding 2 deployments to a service? wouldn't that cause the service selector to include both deployment labels and break the matching?

packages/cdk8s-plus/src/deployment.ts
eladb: You are right. Let me revise.

eladb: Fixed.

packages/cdk8s-plus/src/service.ts
eladb: I wanted an API that allows me to associate a deployment to an existing service (instead of `expose()` which define a new service). The semantic `addXxx` is like saying "add a deployment to the definition of this service".

However, you are right that adding multiple deployments will basically not work. I guess we can fail `addDeployment()` is there are already selectors defined for this service. What do you think?

packages/cdk8s-plus/src/service.ts
eladb: Fixed.

packages/cdk8s-plus/src/service.ts
iliapolo: > The semantic addXxx is like saying "add a deployment to the definition of this service"

Problem is that a deployment doesn't really appear in the spec of a service, so i'm not sure its the best semantic. 
I wonder if maybe `service.serve` might be better. i.e, "this service will serve this deployment" (failing when its called twice) - And change the current `serve` to `bind`? 

Granted it doesn't follow the "add to spec definition" semantic but i dont feel like we have clear convention here..

iliapolo: Btw - we can also just have the `deployment.expose` method accept an optional service

eladb: We are getting rid of specs anyway, no? I can move this method from `ServiceSpec` to `Service`. Would that make more sense?

eladb: done

iliapolo (approved):
I dont want to get hung up on this now, but i'm taking a mental note to revisit the `addDeployment` name :)

eladb:
> I dont want to get hung up on this now, but i'm taking a mental note to revisit the `addDeployment` name :)

Yes I am not in love with this name

PR #346, opened 2020-10-16
feat(lib): helm construct

### LINKED ISSUE #65: Import Helm Charts

mbonig:
Closes #65

This introduces a new construct that can be used to include an existing Helm chart into your constructs.


mbonig:
Build is failing and I'm not sure the best way to fix it. Right now it's failing because running the locally installed helm is failing. I don't see any reason it should be, unless there is a reason the helm install I do earlier in the build doesn't carry over to the build step. I'll keep digging but wouldn't mind any tips on what I'm doing wrong here...

pgollucci:
commenting so I'm in the loop

pgollucci:
Is helm going in /usr/local/bin?  When you run the commands your init files might not be adding it to the path.

The alpine/helm:latest is a whole other thing.  That should exist.

pgollucci:
> Build is failing and I'm not sure the best way to fix it. Right now it's failing because running the locally installed helm is failing. I don't see any reason it should be, unless there is a reason the helm install I do earlier in the build doesn't carry over to the build step. I'll keep digging but wouldn't mind any tips on what I'm doing wrong here...

That could be a docker thing.

mbonig:
Alright, ignore all of that, I got ahead of myself and missed committing a test chart for the unit tests. Still got a docker problem but I think I know what's happening there.

mbonig:
So, I believe the only remaining failure is a docker-in-docker problem. Going to take some queues from the NodejsFunction to fix this.

wuqunfei:
+1

eladb (approved):

guan840912:
enable in 0.34.0 release ?!

PR #347, opened 2020-10-17
refactor(plus): Remove the `spec` nesting level on both input and output

iliapolo:
Removed the need to specify various `spec` properties for both pre and post instantiation. Applies to all constructs that used to accept a pod spec.

### Before

```ts
const deployment = new kplus.Deployment(this, text, {
  spec: {
    podSpecTemplate: {
      containers: [
        new kplus.Container({
          image: 'hashicorp/http-echo',
          args: [ '-text', text ]
        })
      ]
    }
  }
});
deployment.spec.podSpecTemplate.addContainer(...)
```

### After

```ts
const deployment = new kplus.Deployment(this, text, {
  containers: [
    new kplus.Container({
      image: 'hashicorp/http-echo',
      args: [ '-text', text ]
    })
  ]
});
deployment.addContainer(...)
```

BREAKING CHANGE: `spec` was removed from all cdk8s+ constructs and that now have a flat structure. See [Example](https://github.com/awslabs/cdk8s/tree/master/packages/cdk8s-plus#at-a-glance) for new usage.

* **plus**: Construct id's for deployment will change due to a latent bug that appended the word `pod` to them.


packages/cdk8s-plus/src/deployment.ts
```diff
@@ -186,27 +149,42 @@ export class DeploymentSpecDefinition {
...
    * @internal
    */
   public _toKube(): k8s.DeploymentSpec {
     return {
       replicas: this.replicas,
-      template: {
-        metadata: this.podMetadataTemplate.toJson(),
-        spec: this.podSpecTemplate._toKube(),
-      },
+      template: this._podTemplate._toPodTemplateSpec(),
```
eladb, line 183: toKube maybe?

eladb (dismissed):

packages/cdk8s-plus/src/pod.ts
```diff
@@ -5,52 +5,173 @@ import * as cdk8s from 'cdk8s';
...
+
+  constructor(props: PodTemplateProps = {}) {
+    super(props);
+    this.podMetadata = new ApiObjectMetadataDefinition(props.podMetadata);
+  }
+
+  /**
+   * @internal
+   */
+  public _toPodTemplateSpec(): k8s.PodTemplateSpec {
```
iliapolo, line 158: This felt right, letting this class be responsible for creating the template spec used by `Job` and `Deployment`. It did mean I have to change the name if we want to preserve a typed return value from these methods.

We can also make this class not extend `PodSpec` but rather compose it and then we can use the normal `_toKube`, but that felt a bit redundant.

packages/cdk8s-plus/src/deployment.ts
iliapolo: I can't :\ 

`_toKube` is already implemented with a different return value in `PodSpec`. See https://github.com/awslabs/cdk8s/pull/347#discussion_r507086143

iliapolo: oh you mean just without the underscore?

packages/cdk8s-plus/src/pod.ts
eladb: Makes me think we need to rename all these toKube methods to their respective output type

eladb (dismissed):

eladb (approved):

PR #349, opened 2020-10-19
fix(lib): fail to import octal numbers via include (and helm)

### LINKED ISSUE #348: Octal numbers are not parsed correctly when including manifests (or helm charts)
### Description of the bug:

See https://github.com/eemeli/yaml/issues/205

### Reproduction Steps:

Try to `Include` a manifest that includes an octal number:

```yaml
my_number: 0666
```

The expectation is that this will be represented as:

```json
{ "my_number": 438 }
```

But instead, it is represented as `666`, which is not correct.

Root cause is an issue in the YAML library we use: https://github.com/eemeli/yaml/issues/205

### Environment:

  - **Framework Version**:
  - **OS**:

### Other:

--- 

This is :bug: Bug Report

eladb:
When parsing YAML, use yaml-1.1 instead of 1.2 in order to allow octal numbers in the form `0775` (instead of `0o755`, which is what 1.2 requires). See https://github.com/eemeli/yaml/issues/205.

Fixes #348

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,8 @@
     "nohoist": [
       "cdk8s/yaml",
       "cdk8s/yaml/**",
+      "cdk8s/js-yaml",
+      "cdk8s/js-yaml/**",
       "cdk8s/json-stable-stringify",
       "cdk8s/json-stable-stringify/**",
       "cdk8s/follow-redirects",
--- a/package.json
+++ b/package.json
@@ -20,8 +20,6 @@
     "nohoist": [
       "cdk8s/yaml",
       "cdk8s/yaml/**",
-      "cdk8s/js-yaml",
-      "cdk8s/js-yaml/**",
       "cdk8s/json-stable-stringify",
       "cdk8s/json-stable-stringify/**",
       "cdk8s/follow-redirects",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -20,6 +20,8 @@
     "nohoist": [
       "cdk8s/yaml",
       "cdk8s/yaml/**",
+      "cdk8s/js-yaml",
+      "cdk8s/js-yaml/**",
       "cdk8s/json-stable-stringify",
```
eladb, line 25:
```suggestion
      "cdk8s/yaml/**",
      "cdk8s/json-stable-stringify",
```

packages/cdk8s/test/yaml.test.ts
```diff
@@ -77,4 +79,11 @@ describe('save', () => {
     expect(fs.readFileSync(outputFile, 'utf-8')).toMatchSnapshot();
   });
 
-})
+});
+test('yaml 1.1 octal numbers are parsed correctly', () => {
```
eladb, line 83:
```suggestion
});

test('yaml 1.1 octal numbers are parsed correctly', () => {
```

iliapolo (approved):

eemeli:
Just to be clear, with this change the parsing of other values in addition to octal numbers is also changed to follow the much more lax YAML 1.1 rules, such as `no` now being parsed as `false`. I hope that's intentional?

PR #351, opened 2020-10-19
feat(lib): SecretValue

eladb:
Introduce a struct called `SecretValue` which represents a value within a JSON secret. This allow encapsulating a specific secret value into a strongly-typed value.

Misc: make `resolve()` internal by moving to a non-exported file since jsii does not allow free floating functions.

BREAKING CHANGE: `EnvValue.fromSecret(secret, key)` has been removed in favor of `EnvValue.fromSecretValue({ secret, key })`.


packages/cdk8s-plus/src/container.ts
```diff
@@ -70,18 +70,27 @@ export class EnvValue {
...
 
   /**
-   * Create a by reading a specific key inside a secret.
+   * Create an env value by reading a specific key inside a secret.
    *
    * @param secret - The secret.
    * @param key - The key.
    * @param options - Additional options.
    */
   public static fromSecret(secret: ISecret, key: string, options: EnvValueFromSecretOptions = {}): EnvValue {
```
iliapolo, line 79: I'd say we get rid of this to reduce cognitive load, users can just as easily call `fromSecretValue` themselves.

eladb: Okay, fair enough

eladb: fixed and added BREAKING CHANGE

iliapolo (dismissed):

PR #357, opened 2020-10-23
feat(plus): expose service options in `expose()`

abierbaum:
This PR expands the `Deployment expose()` method to allow users to fully specify options to the generated Service and Port.  Originally I created this so I could provide a user defined name for the Service through metadata.  In the process I also found the need to allow specifying the port details for name and protocol.

### DIFF
--- a/packages/cdk8s-plus/src/deployment.ts
+++ b/packages/cdk8s-plus/src/deployment.ts
@@ -47,6 +47,8 @@ export interface ExposeOptions {
   /**
    * The name of the service to expose.
    * This will be set on the Service.metadata and must be a DNS_LABEL
+   * 
+   * @default undefined Uses the system generated name.
    */
   readonly name?: string;
   
@@ -179,7 +181,7 @@ export class Deployment extends Resource implements IPodTemplate {
    */
   public expose(port: number, options: ExposeOptions = {}): Service {
     const service = new Service(this, 'Service', {
-      metadata: options.name == null ? {} : {name: options.name},
+      metadata: options.name ? { name: options.name } : undefined,
       type: options.serviceType ?? ServiceType.CLUSTER_IP,
     });
 
--- a/packages/cdk8s-plus/src/deployment.ts
+++ b/packages/cdk8s-plus/src/deployment.ts
@@ -1,6 +1,6 @@
 import * as k8s from './imports/k8s';
 import { Construct, Node } from 'constructs';
-import { Service, ServiceType } from './service';
+import { Service, ServicePortOptions, ServiceProps, ServiceType } from './service';
 import { Resource, ResourceProps } from './base';
 import * as cdk8s from 'cdk8s';
 import { ApiObjectMetadataDefinition, Names } from 'cdk8s';
@@ -31,18 +31,6 @@ export interface DeploymentProps extends ResourceProps, PodTemplateProps {
    */
   readonly defaultSelector?: boolean;
 
-}
-
-/**
- * Options for exposing a deployment via a service.
- */
-export interface ExposeOptions {
-  /**
-   * The type of the exposed service.
-   *
-   * @default - ClusterIP.
-   */
-  readonly serviceType?: ServiceType;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


abierbaum:
From discussion on slack with: @mbonig 

Any guidance on the best way to expose this API.  I thought more about what you mentioned which I would summarize as extending ExposeOptions with the most common options, but I think the part I run into in my case is that there is no escape hatch that allows direct access to the wrapped api objects.

I could make ExposeOptions something like: 

```
export interface ExposeOptions {
    serviceType?: ServiceType;
    serviceName?: string;
    protocol?: Protocol;
}
```

If someone wanted to got directly at the service and ports though they would only be able to do it through Service.ports and then directly modify the port objects in the list.  If someone wanted to modify the metadata on the Service itself they could always use Service.metadata() to directly access the apiObject.metadata.

Any thoughts @eladb  @iliapolo @mbonig ?  I would love to get this through, just want to make sure to follow any best practices or conventions you have for cdk8s-plus.  It is an awesome project.

abierbaum:
I just updated the API based upon @mbonig feedback.  This makes the most common case pretty easy to use but still allows the use of the escape hatch of Service.metadata and Service.ports.

Please let me know what you think.

packages/cdk8s-plus/src/deployment.ts
```diff
@@ -43,8 +43,29 @@ export interface ExposeOptions {
    * @default - ClusterIP.
    */
   readonly serviceType?: ServiceType;
+
+  /**
+   * The name of the service to expose.
+   * This will be set on the Service.metadata and must be a DNS_LABEL
+   */
```
eladb, line 50: All optional fields must have a "@default" annotation

packages/cdk8s-plus/src/deployment.ts
```diff
@@ -154,14 +175,15 @@ export class Deployment extends Resource implements IPodTemplate {
    * This is equivalent to running `kubectl expose deployment <deployment-name>`.
    *
    * @param port The port number the service will bind to.
-   * @param options Options.
+   * @param options Options to determine details of the service and port exposed.   
    */
   public expose(port: number, options: ExposeOptions = {}): Service {
     const service = new Service(this, 'Service', {
+      metadata: options.name == null ? {} : {name: options.name},
```
eladb, line 182:
```suggestion
      metadata: options.name ? { name: options.name } : undefined,
```

packages/cdk8s-plus/src/service.ts
```diff
@@ -168,14 +168,15 @@ export class Service extends Resource {
...
+   * If not targetPort is specific in the portOptions, then requests will be routed
+   * to the port exposed by the first container in the deployment's pods. 
+   * The deployment's `labelSelector` will be used to select pods.
    *
    * @param deployment The deployment to expose
    * @param port The external port
+   * @param portOptions Optional settings for the port.
    */
-  public addDeployment(deployment: Deployment, port: number) {
+  public addDeployment(deployment: Deployment, port: number, portOptions: ServicePortOptions = {}) {
```
eladb, line 179:
```suggestion
  public addDeployment(deployment: Deployment, port: number, options: ServiceDeploymentOptions = {}) {
```

eladb (changes requested):

packages/cdk8s-plus/src/service.ts
abierbaum: @eladb I can't find ServiceDeploymentOptions and where these options get applied is a call to this.serve(port, {..options}) that only takes a ServicePortOptions.  I am changing the name, but may need a bit more of a pointer to understand the new type.

packages/cdk8s-plus/src/deployment.ts
```diff
@@ -154,14 +142,16 @@ export class Deployment extends Resource implements IPodTemplate {
...
-   * @param options Options.
+   * @param srvOptions Options to configure the wrapped Service.
+   * @param portOptions Optional settings for the exposed port.
    */
-  public expose(port: number, options: ExposeOptions = {}): Service {
+  public expose(port: number, srvOptions: ServiceProps = {}, portOptions: ServicePortOptions = {}): Service {
     const service = new Service(this, 'Service', {
-      type: options.serviceType ?? ServiceType.CLUSTER_IP,
+      ...srvOptions,
+      type: srvOptions.type ?? ServiceType.CLUSTER_IP,
```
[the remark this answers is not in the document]
abierbaum: This line doesn't exist anymore.

packages/cdk8s-plus/src/deployment.ts
abierbaum: I added a default.

packages/cdk8s-plus/src/service.ts
abierbaum: I have renamed options.

abierbaum:
@eladb I have updated the PR based upon your review.  Thanks!!

eladb (approved):

PR #359, opened 2020-10-26
feat(plus): add method to mount raw volume to a container

abierbaum:
This is the best way I have found to use kplus and allow mounting to all the various k8s volume types.

### DIFF
--- a/packages/cdk8s-plus/src/volume.ts
+++ b/packages/cdk8s-plus/src/volume.ts
@@ -99,6 +99,20 @@ export class Volume {
   }
 
   /**
+   * Create a volume the wraps a k8s.Volume for usage in cdk8s-plus.
+   * 
+   * @param name Name of the volume.
+   * @param volume The k8s.Volume resource to wrap.
+   * 
+   * note: 
+   *    Using Omit outputs: 
+   *      jsii/compiler - Type model errors prevented the JSII assembly from being created
+   */
+  public static fromVolumeSpec(name: string, volume: any /*Omit<k8s.Volume, 'name'>*/): Volume {
+    return new Volume(name, volume);
+  }
+
+  /**
    * @internal
    */
   public _toKube(): k8s.Volume {
--- a/packages/cdk8s-plus/API.md
+++ b/packages/cdk8s-plus/API.md
@@ -1733,6 +1733,20 @@ static fromEmptyDir(name: string, options?: EmptyDirVolumeOptions): Volume
 __Returns__:
 * <code>[Volume](#cdk8s-plus-volume)</code>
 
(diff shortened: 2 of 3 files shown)


packages/cdk8s-plus/src/volume.ts
```diff
@@ -98,6 +98,20 @@ export class Volume {
...
+   * Create a volume the wraps a k8s.Volume for usage in cdk8s-plus.
+   * 
+   * @param name Name of the volume.
+   * @param volume The k8s.Volume resource to wrap.
+   * 
+   * note: 
+   *    Using Omit outputs: 
+   *      jsii/compiler - Type model errors prevented the JSII assembly from being created
+   */
+  public static fromVolumeSpec(name: string, volume: any /*Omit<k8s.Volume, 'name'>*/): Volume {
```
abierbaum, line 111: I would love to find a way to use Omit here but it causes an error with the jsii/compiler:

```
jsii/compiler - Type model errors prevented the JSII assembly from being created
```

Any ideas?

abierbaum:
@iliapolo Any thoughts on this?  If it isn't something the team wants, I can extract it to an addon in my local project.  Just want to know if I should close it.

iliapolo:
@abierbaum Sorry for the delay. 

I understand the use case of this PR, but we deliberately try not to expose the underlying L1 spec. 
So I feel this needs a different approach.

I'm ok with leaving it open for now just so we will have it on the radar for later. Added the `needs-discussion` label to signal that.

abierbaum:
@iliapolo I don't disagree that it is ugly and should be improved.  In the meantime, what is your recommended way to do something like add a host volume.  I only added this path because I couldn't see another way to do it.

iliapolo:
Hi @abierbaum - Apologies for the delayed response. I think the best option right now is to use escape hatches, which we've just recently added:

https://cdk8s.io/docs/v1.0.0-beta.3/concepts/escape-hatches/

Going forward i think we should add `Volume.fromHostPath` API, similar to like the ones we already have. WDYT? 

Thanks

abierbaum:
> Hi @abierbaum - Apologies for the delayed response. I think the best option right now is to use escape hatches, which we've just recently added:
> 
> https://cdk8s.io/docs/v1.0.0-beta.3/concepts/escape-hatches/
> 
> Going forward i think we should add `Volume.fromHostPath` API, similar to like the ones we already have. WDYT?
> 
> Thanks

Seems reasonable.  One thing I would suggest considering is exposing the Volume constructor to allow more directly use when requested.  The JSONPatch support could be used, but seems like it would be a bit of a pain to do things that are abnormal.

For my cases I just created a custom construct that extends Volume and supports the use cases I need.

PR #370, opened 2020-11-10
feat(cli): class prefix for imported constructs

### LINKED ISSUE #140: [cdk8s-cli] Incorrect manifest when configuring a `StatefulSet` with `volumeClaimTemplates`
**Describe the bug**

A `StatefulSet` that defines `VolumeClaimTemplates` is incorrectly synthesized.

**To Reproduce**

```typescript
new k8s.StatefulSet(this, 'StatefulSet', {
  spec: {
    serviceName: 'service',
    selector: {
      matchLabels: { a: 'b' }
    },
    template: {
      spec: {
        containers: [{
          name: 'hello'
        }]
      }
    },
    volumeClaimTemplates: [new k8s.PersistentVolumeClaim(this, 'VolumeClaim', {
      spec: {
        accessModes: ['ReadWriteOnce']
      },
    })]
  }
})
```

This will produce the following manifest:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: cdk8s-playground-volumeclaim-0db7a75b
spec:
  accessModes:
    - ReadWriteOnce
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  clusterName: cluster
  name: cdk8s-playground-statefulset-4c20ee82
spec:
  selector:
    matchLabels:
      a: b
  serviceName: service
  template:
    spec:
      containers:
        - name: hello
  volumeClaimTemplates:
    - apiVersion: v1
      chart:
        manifestFile: cdk8splayground.k8s.yaml
      kind: PersistentVolumeClaim
      name: cdk8s-playground-volumeclaim-0db7a75b
      options:
        apiVersion: v1
        kind: PersistentVolumeClaim
        spec:
          accessModes:
            - ReadWriteOnce
```

There are two problems here:

1. `PersistentVolumeClaim` is present both as a top level resource, **and** a nested resource inside the spec of the`StatefulSet`.

2. The nested resource spec contains keys that actually make it invalid. For example, `chart`, `options`...

**Expected behavior**

According to the kubernetes [spec](https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.15.0/_definitions.json#L1178) for a `StatefulSet`, I would expect the synthesized manifest to look like so:

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  clusterName: cluster
  name: cdk8s-playground-statefulset-4c20ee82
spec:
  selector:
    matchLabels:
      a: b
  serviceName: service
  template:
    spec:
      containers:
        - name: hello
  volumeClaimTemplates:
    - apiVersion: v1
      kind: PersistentVolumeClaim
      metadata:
        name: cdk8s-playground-volumeclaim-0db7a75b
      spec:
        accessModes:
          - ReadWriteOnce
``` 

**Additional context**

The problem stems from the fact that the [`volumeClaimTemplates`](https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.15.0/_definitions.json#L1178) property of the `StatefulSet` spec [references](https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.15.0/_definitions.json#L1214) a [`PersistentVolumeClaim`](https://github.com/instrumenta/kubernetes-json-schema/blob/master/v1.15.0/_definitions.json#L8886) object, which is, in a very non characteristic way, an `ApiObject` itself, rather than a simple struct.

The in turn causes the generated code to force users to create a top level `PersistentVolumeClaim` object:

```typescript
volumeClaimTemplates: [new k8s.PersistentVolumeClaim(this, 'VolumeClaim', {
      spec: {
        accessModes: ['ReadWriteOnce']
      },
    })]
```

Which in turn causes all the problems mentioned here.

As a side thought, even if the synthesis was somehow correct, the API presented here is wrong and doesn't correctly represent the mental model of the use-case, as there is no reason to create a `PersistentVolumeClaim` construct (resource), if all thats needed is to pass its spec to a parent resource. 

This implies that the generated code has to change has to change, and provide a way to simply specify the spec for the `volumeClaimTemplates` property, and not the actual construct.

eladb:
Adds a new `--class-prefix` option to `cdk8s import` which will prepend all construct classes with the specified prefix. By default, `k8s` imports will now have a `Kube` prefix in order to differentiate them from the high-level APIs in CDK8s+. CRD imports will not have a prefix.

This is in-line with how L1s are imported in the AWS CDK (they all have a `Cfn` prefix).

Indirectly resolves #140 - since now imported k8s class names will have a different name from struct names, so there will be no conflict when `PersistentVolumeClaim` is used as a data type and not as a resource (inside `StatefulSetSpec`).

NOTE: Since the core Kubernetes API sometimes references API objects from the specs of other API objects (e.g. the `items` field of `DeploymentList` refers to `Deployment`), we will use the `KubeDeploymentProps` type instead in that case.

BREAKING CHANGE: When using the CLI to import the core Kubernetes API objects, the imported classes will now have a `Kube` prefix in order to make it easier to differentiate them from the classes offered by the high-level APIs in CDK8s+ (e.g. `k8s.Deployment` is now `k8s.KubeDeployment`). You can disable through the `--no-class-prefix` option: `cdk8s import --no-class-prefix k8s`.

---

### TODO

- [x] Docs
- [x] Examples
- [x] Integration tests
- [x] CDK8s+


packages/cdk8s-plus/package.json
```diff
@@ -65,14 +66,72 @@
...
     "k8s",
-    "constructs"
+    "kubernetes"
   ],
   "engines": {
     "node": ">= 10.17.0"
   },
+  "main": "lib/index.js",
   "license": "Apache-2.0",
+  "start": {
```
iliapolo, line 78: Just out of curiosity - what's this?

packages/cdk8s-plus/src/imports/k8s.ts
```diff
@@ -5475,6 +5592,28 @@ export interface ReplicaSetSpec {
 
 }
 
+/**
+ * ReplicaSet ensures that a specified number of pod replicas are running at any given time.
+ *
+ * @schema io.k8s.api.apps.v1.ReplicaSet
+ */
+export interface ReplicaSet {
```
iliapolo, line 5600: Ha - since now the generated constructs don't register with any existing names, we have duplicate interfaces (i.e `ReplicateSet` and `RepliateSetOptions`).

As we discussed, seems like a bug

iliapolo (changes requested):
Just the duplicate interfaces thingy. Looks great otherwise.

packages/cdk8s-plus/package.json
eladb: It's the metadata for `projen start` (added in the projen upgrade).

packages/cdk8s-plus/src/imports/k8s.ts
eladb: Fixed.

iliapolo (approved):

PR #372, opened 2020-11-12
feat: escape hatches

### LINKED ISSUE #144: [cdk8s] Escape hatches
Similarly to how [escape hatches] work in the AWS CDK.

[escape hatches]: https://docs.aws.amazon.com/cdk/latest/guide/cfn_layer.html

eladb:
Introduce `apiObject.addJsonPatch(op)` as a way to patch the synthesized manifest of an object.

Introduce `ApiObject.of(c)` which returns the first default child of type `ApiObject` as a way to obtain the underlying L1 from a higher-level construct.

Example:

    ApiObject.of(pod).addJsonPatch(
      JsonPatch.add('/spec/enableServiceLinks', true));

Resolves #144


packages/cdk8s/src/api-object.ts
```diff
@@ -34,6 +35,27 @@ export interface ApiObjectOptions {
 }
 
 export class ApiObject extends Construct {
+  /**
+   * Returns the `ApiObject` named `Default` which is a child of the given
```
iliapolo, line 39: Should this be changed to:

```suggestion
   * Returns the `ApiObject` named `Resource` which is a child of the given
```

?

https://github.com/awslabs/cdk8s/issues/373

packages/cdk8s/src/api-object.ts
```diff
@@ -34,6 +35,27 @@ export interface ApiObjectOptions {
...
+   * @param c The higher-level construct
+   */
+  public static of(c: IConstruct): ApiObject {
+    if (c instanceof ApiObject) {
+      return c;
+    }
+
+    const child = Node.of(c).defaultChild;
+    if (!child) {
+      throw new Error('cannot find a (direct or indirect) child of type ApiObject');
```
iliapolo, line 53:
```suggestion
      throw new Error(`cannot find a (direct or indirect) child of type ApiObject for construct ${Node.of(c).addr}`);
```

iliapolo (approved):

PR #376, opened 2020-11-13
feat(plus): Container is now inputed as an interface instead of class

iliapolo:
When a container needs to be passed as an input, it can be passed with a simple interface.

```ts
const pod = new kplus.Pod(this, 'Pod', {
  containers: [{ image: 'nginx' }]
});
```

When a container is retrieved as an output, it is a class, with the familiar api.

```ts
const container = pod.addContainer({ image: 'nginx' });
container.mount(...)
```

BREAKING CHANGE: Containers now need to be inputed as interfaces rather than classes. Instead of passing `new kplus.Container(props)`, simply pass in `props`.

Resolves https://github.com/awslabs/cdk8s/issues/365


examples/typescript/cdk8s-plus-elasticsearch-query/__snapshots__/main.test.ts.snap
```diff
@@ -42,82 +42,6 @@ Array [
       "version": "7.7.1",
     },
   },
-  Object {
```
iliapolo, line 45: Not sure why this changed locations...but it looks ok.

packages/cdk8s-plus/src/pod.ts
```diff
@@ -82,7 +82,7 @@ export class PodSpec implements IPodSpec {
     this.restartPolicy = props.restartPolicy;
     this.serviceAccount = props.serviceAccount;
 
-    this._containers = props.containers ?? [];
+    this._containers = props.containers?.map(c => new Container(c)) ?? [];
```
eladb, line 85: Worth calling `addContainer()` instead of reimplementing it so mutating the internal data structure happens in one place.

eladb (approved):

PR #379, opened 2020-11-16
feat(cli): import constructs for all API levels

### LINKED ISSUE #380: Annotate non-stable API levels in imported k8s resources
Instead of filtering out non-stable API resources in `cdk8s import k8s`, we should simply append an API level annotation to those classes. For example, in 1.18, it should be `IngressV1Beta1` instead of just `Ingress`. This is more in-line with how k8s API levels work. Before a resource is stable, breaking changes can be introduced. So we want to make sure users (and cdk8s-plus) are more explicit about their usage of non-stable resources.

eladb:
Following this change, when using `cdk8s import` to generate constructs from schemas, the CLI will generate constructs for all API levels and not just the latest stable version.

The name of the generated class will now include the API level, unless this is a `v1` resource. For example, `v1/Pod` will be called `Pod`, while `v1beta1/Ingress` will be called `IngressV1Beta1`.

This change only applies to `cdk8s import k8s` and does not apply when importing CRDs, in which case a single CRD always includes a single resource definition.

BREAKING CHANGE: when importing k8s api objects using `cdk8s import`, non-stable APIs will be have an API level postfix. For example, k8s@1.18 will have an `IngressV1Beta1` API object.
* **cli:** The `--include` CLI option has been removed since all API objects are always imported.

Resolves #380

---


packages/cdk8s-cli/src/cli/cmds/import.ts
```diff
@@ -20,7 +20,6 @@ class Command implements yargs.CommandModule {
     .example('cdk8s import mattermost:=mattermost_crd.yaml', 'Imports constructs for the mattermost cluster custom resource definition using a custom module name')
 
     .option('output', { default: DEFAULT_OUTDIR, type: 'string', desc: 'Output directory', alias: 'o' })
-    .option('include', { type: 'array', desc: 'Types to select instead of the default which is latest stable version (only for "k8s")' })
```
iliapolo, line 23: Mention this removal in the changelog and docs.

iliapolo (changes requested):
Also need to re-import cdk8s+

iliapolo (approved):

packages/cdk8s-cli/src/cli/cmds/import.ts
eladb: done

PR #389, opened 2020-11-18
fix(plus-17): L2 default child

iliapolo:
All L1 children are now called `Resource` to mark them as the default child of the L2.

BREAKING CHANGE: All L2 resource names will undergo a name change (e.g `test-chart-config-configmap-233db8e7` -> `test-chart-config-c3f7d3c0`)

Resolves https://github.com/awslabs/cdk8s/issues/373


packages/cdk8s-plus-17/src/config-map.ts
```diff
@@ -67,7 +67,7 @@ export class ConfigMap extends Resource implements IConfigMap {
   public constructor(scope: Construct, id: string, props: ConfigMapProps = { }) {
     super(scope, id, { metadata: props.metadata });
 
-    this.apiObject = new k8s.KubeConfigMap(this, 'ConfigMap', {
+    this.apiObject = new k8s.KubeConfigMap(this, 'Resource', {
```
eladb, line 70: We should add `Construct.DEFAULT_CHILD` or something like that, in the future.

packages/cdk8s-plus-17/test/config-map.test.ts
```diff
@@ -1,6 +1,15 @@
 import { Testing } from 'cdk8s';
+import { Node } from 'constructs';
 import { ConfigMap } from '../src';
 
+test('defaultChild', () => {
+
+  const chart = Testing.chart();
+
+  expect(Node.of(new ConfigMap(chart, 'ConfigMap')).defaultChild).toBeTruthy();
```
eladb, line 9: would it be better to assert `instanceof KubeConfigMap` or at least `ApiObject`.

BTW, should we rename `ApiObject` to `KubeObject`?

packages/cdk8s-plus-17/test/deployment.test.ts
```diff
@@ -1,6 +1,15 @@
 import { Testing } from 'cdk8s';
+import { Node } from 'constructs';
 import * as kplus from '../src';
 
+test('defaultChild', () => {
```
eladb, line 5: how do we enforce that such a test exists for all L2s? (for later)

packages/cdk8s/src/names.ts
```diff
@@ -142,6 +122,26 @@ function omitDuplicates(value: string, index: number, components: string[]) {
   return value !== components[index-1];
 }
 
+function omitDefaultChild(value: string, _: number, __: string[]) {
+  return value.toLowerCase() !== 'resource' && value.toLowerCase() !== 'default';
```
eladb, line 126: I believe it is actually case sensitive in `constructs`.

packages/cdk8s/src/names.ts
```diff
@@ -53,16 +53,7 @@ export class Names {
 
     components.push(calcHash(path, HASH_LEN));
```
eladb, line 54: We should use `Node.of(c).addr` instead of manually calculating the hash here.

eladb (changes requested):

packages/cdk8s-plus-17/test/deployment.test.ts
iliapolo: Marked for later

eladb (approved):

packages/cdk8s/src/names.ts
iliapolo: Noted

PR #396, opened 2020-11-19
feat(lib): allow hash to be optionally included in Names functions.

abierbaum:
This makes it easier for users that want to use custom names without hashes.

Note: it would also be helpful if the functions `toHumanForm`, `normalizeToDnsName`, and `normalizeToLabelValue` were exposed.  This would let them be reused if users want completely customized names.


eladb:
Not sure I fully understand the use case. Without hashes the auto-generated names can conflict quite easily. If you want to control the name of a resource, you can just assign it directly through `metadata.name`. Isn't that sufficient?

abierbaum:
@eladb I could be missing something.  In our usecase we don't run into naming conflicts, so having the stable names was more attractive to us.  We create deployments, stateful sets, services, etc.  Then the sub-resources allocated by those end up with the normal k8s unique names.  But we want to have stable names for the top level resources.

Originally I had simply copied your code into another file and override the `generateObjectName` method.  This works, but I was trying to find a way to not have to copy all of the cdk8s name generation code to simply remove the hash.  I am fine with doing that though if this isn't something the project wants to expose.

eladb (approved):

PR #398, opened 2020-11-19
feat(plus-17): additional options for the Job construct.

abierbaum:


abierbaum:
@mergifyio refresh

packages/cdk8s-plus-17/src/job.ts
```diff
@@ -13,6 +13,40 @@ import { Volume } from './volume';
  */
 export interface JobProps extends ResourceProps, PodTemplateProps {
 
+  /**
+   * Specifies the duration in seconds relative to the startTime that
+   * the job may be active before the system tries to terminate it.
```
iliapolo, line 18:
```suggestion
   * Specifies the duration in seconds the job may be active before the system tries to terminate it.
```

packages/cdk8s-plus-17/src/job.ts
```diff
@@ -37,6 +71,26 @@ export interface JobProps extends ResourceProps, PodTemplateProps {
...
+
+  /**
+   * Number of retries before marking failed.
+   */
+  public readonly backoffLimit?: number;
+
+  /**
+   * Number of successful completions required for job.
+   */
+  readonly completions?: number;
```
iliapolo, line 87:
```suggestion
public  readonly completions?: number;
```

packages/cdk8s-plus-17/src/job.ts
```diff
@@ -37,6 +71,26 @@ export interface JobProps extends ResourceProps, PodTemplateProps {
...
+
+  /**
+   * Max pods to run at a given time.
+   */
+  readonly parallelism?: number;
```
iliapolo, line 92:
```suggestion
public  readonly parallelism?: number;
```

packages/cdk8s-plus-17/src/job.ts
```diff
@@ -13,6 +13,40 @@ import { Volume } from './volume';
...
+
+  /**
+   * Specifies the maximum desired number of pods the job should run at any given time.
+   * The actual number of pods running in steady state will be less than this number when
+   * ((.spec.completions - .status.successful) < .spec.parallelism),
+   * i.e. when the work left to do is less than max parallelism.
+   *
+   * @defaul - If unset, there will be no parallelism.
+   */
+  readonly parallelism?: number;
```
iliapolo, line 48: If there is no parallelism, is there meaning to the `completions` property? i.e does specifying 2 completions with no parallelism makes sense? 

Im wondering what the relationship is between these two properties and if we can somehow combine them to something less error prone.

packages/cdk8s-plus-17/src/job.ts
```diff
@@ -13,6 +13,40 @@ import { Volume } from './volume';
...
+  /**
+   * Specifies the number of retries before marking this job failed.
+   *
+   * @default - If not set, system defaults to 6.
+   */
+  readonly backoffLimit?: number;
+
+  /**
+   * Specifies the desired number of successfully finished pods the job should be run with.
+   * Setting to 1 means that parallelism is limited to 1
```
iliapolo, line 33: I'm not sure I understand this. Does this also mean that setting it to 2 will limit parallelism to 2? 
Intuitively i imagined that it must be that `spec.completions <= spec.parallelism`

packages/cdk8s-plus-17/test/job.test.ts
```diff
@@ -12,6 +12,33 @@ test('defaultChild', () => {
 
 });
 
+test('Allows setting all options', () => {
+
+  const chart = Testing.chart();
+
+  const job = new kplus.Job(chart, 'Job', {
+    containers: [new kplus.Container({ image: 'image' })],
```
iliapolo, line 20:
```suggestion
    containers: [{ image: 'image' }],
```

https://github.com/awslabs/cdk8s/pull/404 :)

packages/cdk8s-plus-17/test/job.test.ts
```diff
@@ -12,6 +12,33 @@ test('defaultChild', () => {
...
+  const spec = Testing.synth(chart)[0].spec;
+
+  expect(spec.activeDeadlineSeconds).toEqual(20);
+  expect(spec.backoffLimit).toEqual(job.backoffLimit);
+  expect(spec.completions).toEqual(job.completions);
+  expect(spec.parallelism).toEqual(job.parallelism);
+  expect(spec.ttlSecondsAfterFinished).toEqual(1);
+
+  // assert the job object has it.
+  expect(job.restartPolicy).toEqual(kplus.RestartPolicy.NEVER);
```
iliapolo, line 38: assert on the new properties as well.

packages/cdk8s-plus-17/src/job.ts
abierbaum: This part always seemed a bit confusing to me as well.  I am simply copying the documentation for the properties of completions and parallelism from https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#jobspec-v1-batch.  I think the best reference is what they refer to from there about https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

packages/cdk8s-plus-17/test/job.test.ts
abierbaum: Just pushed up.

abierbaum:
@iliapolo I think I have addressed all the comments in your review.  Thanks for taking a look.

packages/cdk8s-plus-17/src/job.ts
iliapolo: @abierbaum Yeah its definitely tricky. But I think this requires some more attention.

One of the biggest benefits of `cdk8s-plus` is the ability to reduce cognitive load, i.e simplify and clarify confusing properties and features. It's up to us as authors to fully understand how everything works and come up with better modeling. This presents a great opportunity to do that.

The [documentation](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/) you pointed to can be codified, making it so users don't have to go through it to figure out what configuration should be applied to get their desired outcome.

 For example, it states that:

> There are three main types of task suitable to run as a Job:
> 1. Non-parallel Jobs
> 2. Parallel Jobs with a fixed completion count
> 3. Parallel Jobs with a work queue

So we can think of something like:

```ts
// default - non parallel single completion. doesn't offer 'completions` or `parallelism`
new kplus.Job(this, 'Job');

// fixed completion, offers a mandatory completion count, and an optional parallelism that default to 1 and doesn't
// allow a number greater than count
new kplus.FixedCompletionJob(this, 'Job', {
  count: 5
})

// work queue, offers only parallelism as per the docs: 
// For a work queue Job, you must leave .spec.completions unset, and set .spec.parallelism to a non-negative integer.
new kplus.WorkQueueJob(this, 'Job', {
  parallelism: 5
})
```

This can also look something like:

```ts
new kplus.Job(this, 'Job', {
  execution: Execution.fixedCompletion(...) // or Execution.workQueue(...) 
})
```

WDYT? Would you be up for creating a separate PR for the parallelism stuff where we can discuss the right approach?

I'm good with merging the other two properties since they seem to be really straight forward.
 
Also note that since we now have [escape hatches](https://cdk8s.io/docs/v1.0.0-beta.3/concepts/escape-hatches/), while not super elegant, users will always be able to explicitly set any property they want if we haven't yet come up with the right API for it.

abierbaum: @iliapolo I can do that.  May be a couple days.

As to the escape hatches, one question there.  As I used cdk8s+ more, I see places where other properties are missing (for example `Deployment` doesn't expose `strategy` and `revisionHistoryLimit`).  These are to important ones for what I am doing with a deployment.  But what is the general policy about what part of the spec's should get exposed and not?

I understand the idea of reducing cognitive load, and agree that it is a great goal.  cdk8s+ is massively helpful to reduct the cognitive load of selectors and connecting things up like services to deployments. 

I just don't know where a good line is for what is exposed because if something doesn't get exposed to pass to the apiObject spec below it, then suddenly the cognitive load massively increases because people have to start thinking about JSON patching instead of object/interface properties.

How are you thinking about that part?  It would help me understand where to submit PRs for general use or just use json patch for my local use case.

abierbaum:
@iliapolo I have removed the completions and parallel configuration options.

packages/cdk8s-plus-17/src/job.ts
iliapolo: @abierbaum Eventually we'd like cdk8s+ to cover the entire surface area, so there is no line with regards to what should or shouldn't be exposed. 

The only thing to consider is whether the API **can** offer a better experience than just passing through the properties, and if it can, we should hold back until we come up with that API.

Sometimes that won't be necessary, like for the `activeDeadLine` property, and sometimes it won't be possible. But we should ask ourself that question for every property, and whether or not to submit a PR will depend on the answer.

If you find yourself unsure, just open an issue where we can discuss the API prior to diving into the code.

packages/cdk8s-plus-17/src/job.ts
```diff
@@ -13,6 +13,20 @@ import { Volume } from './volume';
  */
 export interface JobProps extends ResourceProps, PodTemplateProps {
 
+  /**
+   * Specifies the duration in seconds the job may be active before the system tries to terminate it.
```
iliapolo, line 17:
```suggestion
   * Specifies the duration the job may be active before the system tries to terminate it.
```

packages/cdk8s-plus-17/src/job.ts
```diff
@@ -100,6 +126,8 @@ export class Job extends Resource implements IPodTemplate {
   public _toKube(): k8s.JobSpec {
     return {
       template: this._podTemplate._toPodTemplateSpec(),
+      activeDeadlineSeconds: this.activeDeadline ? this.activeDeadline?.toSeconds() : undefined,
```
iliapolo, line 129:
```suggestion
      activeDeadlineSeconds: this.activeDeadline?.toSeconds(),
```

packages/cdk8s-plus-17/src/job.ts
```diff
@@ -37,6 +51,16 @@ export interface JobProps extends ResourceProps, PodTemplateProps {
  */
 export class Job extends Resource implements IPodTemplate {
 
+  /**
+   * Duration before job is terminated. (must be integer number of seconds)
+   */
```
iliapolo, line 56:
```suggestion
  /**
   * Duration before job is terminated. If undefined, there is no deadline.
   */
```

packages/cdk8s-plus-17/API.md
```diff
@@ -698,6 +698,8 @@ new Job(scope: Construct, id: string, props?: JobProps)
   * **serviceAccount** (<code>[IServiceAccount](#cdk8s-plus-17-iserviceaccount)</code>)  A service account provides an identity for processes that run in a Pod. __*Default*__: No service account.
   * **volumes** (<code>Array<[Volume](#cdk8s-plus-17-volume)></code>)  List of volumes that can be mounted by containers belonging to the pod. __*Default*__: No volumes.
   * **podMetadata** (<code>[ApiObjectMetadata](#cdk8s-apiobjectmetadata)</code>)  The pod metadata. __*Optional*__
+  * **activeDeadline** (<code>[Duration](#cdk8s-duration)</code>)  Specifies the duration in seconds the job may be active before the system tries to terminate it. __*Default*__: If unset, then there is no deadline.
```
iliapolo, line 701:
```suggestion
  * **activeDeadline** (<code>[Duration](#cdk8s-duration)</code>)  Specifies the duration the job may be active before the system tries to terminate it. __*Default*__: If unset, then there is no deadline.
```

iliapolo (approved):

PR #400, opened 2020-11-19
feat(plus-17): add StatefulSet construct

abierbaum:
This is a first pass at a statefulset construct. It works well for the cases I need and should serve well as a starting point.  Note that I copied the concepts from Deployment so it should look pretty familiar.


abierbaum:
@iliapolo Any thoughts on this one?  I know there could probably be a set of higher level constructs still for some common statefulset cases, but this seems like a pretty good place to start as it basically mirrors Deployment.

iliapolo:
@abierbaum only glanced through it, looks ok, i'll take a closer look soon.

packages/cdk8s-plus-17/src/statefulset.ts
```diff
@@ -0,0 +1,217 @@
...
+    * @default PodManagementPolicy.ORDERED_READY
+    */
+  readonly podManagementPolicy?: PodManagementPolicy;
+
+  /**
+    * Details of service to create and expose.
+    *
+    * @default undefined - allocate a default service based upon servicePort and serviceName.
+    */
+  readonly service?: ServiceProps;
```
iliapolo, line 50: We can just accept a `Service` here instead, it will prevent some logic duplication and keep things more modular. Its ok to require the caller to instantiate other constructs.

packages/cdk8s-plus-17/src/statefulset.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+    * Default port to expose if service doesn't includes ports.
+    *
+    * Must set this or include ports in service props.
+    *
+    * @default undefined
+    */
+  readonly servicePort?: number;
```
iliapolo, line 59: Won't be needed is we accept a `Service`.

packages/cdk8s-plus-17/src/statefulset.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  /**
+    * Name to use for the service to allow finding it in DNS.
+    *
+    * @default undefined
+    */
+  readonly serviceName?: string;
```
iliapolo, line 66: same

packages/cdk8s-plus-17/src/statefulset.ts
```diff
@@ -0,0 +1,217 @@
...
+      this._service.addSelector(k, v);
+    }
+  }
+
+  get podMetadata(): ApiObjectMetadataDefinition {
+    return this._podTemplate.podMetadata;
+  }
+
+  /**
+    * The labels this deployment will match against in order to select pods.
```
iliapolo, line 160:
```suggestion
    * The labels this statefulset will match against in order to select pods.
```

packages/cdk8s-plus-17/src/statefulset.ts
```diff
@@ -0,0 +1,217 @@
...
+    *
+    * Returns a a copy. Use `selectByLabel()` to add labels.
+    */
+  get labelSelector(): Record<string, string> {
```
iliapolo, line 164: To be consistent:

```suggestion
public  get labelSelector(): Record<string, string> {
```

iliapolo (changes requested):

iliapolo:
Dont close

packages/cdk8s-plus-17/src/statefulset.ts
abierbaum: I thought about that but it ends up making the initialization more complex.  I was trying to get an API as easy to use as the deployment expose() call.

For example when I am using this in my production chart I end up calling as:

```
const redis_set = new kplus.StatefulSet(this, 'redis-set', {
         serviceName: 'redis-srv',
         replicas: 2,
         containers: [redis_container],
         servicePort: redis_port,
         service: {
            type: kplus.ServiceType.CLUSTER_IP,
            clusterIP: undefined, // force it to allocate
         },
      });
```

If the user had to allocate the Service it would look like:

```
const redis_srv = new Service(this, 'RedisService', {
   metadata: {name: 'redis-srv'},
   type: ServiceType.CLUSTER_IP,
   clusterIP: undefined,
   ports: [{port: redis_port}],
})

const redis_set = new kplus.StatefulSet(this, 'redis-set', {
         serviceName: 'redis-srv',
         replicas: 2,
         containers: [redis_container],
         servicePort: redis_port,
         service: redis_srv
      });
```
And the callee would have to make sure the service name strings match and the ports match.  Seems like this is the type of thing we would want to wrap to make it simpler to handle and then if they need to modify the service they can patch it.

What do you think though?  I probably missed something in this one.

packages/cdk8s-plus-17/src/statefulset.ts
abierbaum: We would still need it if the Service was optional.  But if the service was required we could remove it yes.

packages/cdk8s-plus-17/src/statefulset.ts
abierbaum: Got it.  I am so used to my production projects just adding this type of thing automatically using prettier and tslint that I forgot.  Sorry.

abierbaum:
@iliapolo Sorry for the delay.  I got this working to where I was able to use it in production and forgot to come back to clean it up for upstream consumption.  I have reviewed your comments, made some fixes, and added some feedback.  Please let me know your thoughts.  I would love to get this through this week.

iliapolo (changes requested):

iliapolo:
Dont close

abierbaum:
@iliapolo I think this last update should address the core issue pretty well.  I changed it to take a Service as input.  After seeing it in action, I agree this is a good change for this use case.

abierbaum:
@iliapolo Have you had a chance to get back to this one yet?

packages/cdk8s-plus-17/src/statefulset.ts
```diff
@@ -0,0 +1,189 @@
...
+import { Resource, ResourceProps } from './base';
+import { Container, ContainerProps } from './container';
+import * as k8s from './imports/k8s';
+import { RestartPolicy, PodTemplate, IPodTemplate, PodTemplateProps } from './pod';
+import { Service } from './service';
+import { IServiceAccount } from './service-account';
+import { Volume } from './volume';
+
+
+export enum PodManagementPolicy {
```
iliapolo, line 12: Add some doc strings. Should also be available in the k8s spec.

packages/cdk8s-plus-17/test/statefulset.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  expect(defaultChild.kind).toEqual('StatefulSet');
+
+});
+
+test('A label selector is automatically allocated', () => {
+
+  const chart = Testing.chart();
+
+  const service = new kplus.Service(chart, 'test_service', {ports: [{port: 80}] });
```
iliapolo, line 20:
```suggestion
  const service = new kplus.Service(chart, 'TestService', {ports: [{port: 80}] });
```

packages/cdk8s-plus-17/test/statefulset.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  const resources = Testing.synth(chart);
+  const srv_spec = resources[0].spec;
+  const set_spec = resources[1].spec;
+  expect(set_spec.replicas).toEqual(1);
+  expect(set_spec.serviceName).toEqual(resources[0].metadata.name);
+  expect(set_spec.podManagementPolicy).toEqual(kplus.PodManagementPolicy.ORDERED_READY);
+
+  expect(srv_spec.type).toEqual(kplus.ServiceType.CLUSTER_IP);
+  expect(srv_spec.clusterIP).toBeUndefined();
+  expect(srv_spec.ports![0].port).toEqual(80);
```
iliapolo, line 106: This doesn't seem related to a `statefulset` test..if its not covered in the `service` tests, lets add it there, but if it is, we can drop it.

packages/cdk8s-plus-17/test/statefulset.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  const service = new kplus.Service(chart, 'test_service', {ports: [{port: 80}] });
+  new kplus.StatefulSet(chart, 'StatefulSet', {
+    containers: [{ image: 'image' }],
+    service,
+  });
+
+  const resources = Testing.synth(chart);
+  const srv_spec = resources[0].spec;
+  const set_spec = resources[1].spec;
```
iliapolo, line 99:
```suggestion
  const setSpec = resources[1].spec;
```

packages/cdk8s-plus-17/test/statefulset.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  expect(set_spec.replicas).toEqual(5);
+  expect(set_spec.serviceName).toEqual('test-srv');
+  expect(set_spec.podManagementPolicy).toEqual(kplus.PodManagementPolicy.PARALLEL);
+
+  expect(srv_spec.type).toEqual(kplus.ServiceType.NODE_PORT);
+  expect(srv_spec.ports![0].port).toEqual(9000);
```
iliapolo, line 134: again, doesn't seem necessary.

iliapolo (changes requested):
@abierbaum this is great. Just some minor nit-picks around naming conventions :)

packages/cdk8s-plus-17/src/statefulset.ts
abierbaum: Updated.

packages/cdk8s-plus-17/test/statefulset.test.ts
abierbaum: Changed.

packages/cdk8s-plus-17/test/statefulset.test.ts
abierbaum: Removed.

packages/cdk8s-plus-17/test/statefulset.test.ts
abierbaum: changed.

packages/cdk8s-plus-17/test/statefulset.test.ts
abierbaum: removed.

abierbaum:
@iliapolo I have made the changes suggested.  I apologize to all for the amount time it took to respond and get everything completed.

iliapolo (approved):

iliapolo:
@abierbaum No worries, i've been lagging as well :)

Thanks so much for this PR, its awesome.

iliapolo (approved):

PR #378, opened 2020-11-16
feat(cli): import from crds.dev

prasek:
Introduce crds.dev import shortcut: `cdk8s import github:account/repo[@VERSION]`

Example:
`cdk8s import github:crossplane/crossplane@0.14.0`

Resolves #377

In a future PR would like to generate a single module per k8s API group with all CRDs for that API group, so you could:
```ts
import { CompositeResourceDefinition, Composition } from './imports/apiextensions.crossplane.io';
```

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

PR #402, opened 2020-11-19
feat(cli): import a single module per api group

### LINKED ISSUE #401: [cdk8s-cli] emit a single module per k8s API group as default for import from crds.dev
Follow on to #377 

Would like to generate a single module per k8s API group with all CRDs for that API group, so you could:
```ts
import { CompositeResourceDefinition, Composition } from './imports/apiextensions.crossplane.io';
```

and have this be the default for importing from https://doc.crds.dev using:
```
cdk8s import github:account/repo[@VERSION]
```

prasek:
Follow-on to #378 

Emit a single module per API group for all imports:
```sh
cdk8s import github:account/repo[@VERSION]
```
you can use a single line import statement for multiple API types:
```ts
import { CompositeResourceDefinition, Composition } from './imports/apiextensions.crossplane.io';
```

Here's some sample output:
https://github.com/prasek/crossplane-cdk/tree/cross-pkg/test/imports/provider-aws

Resolves #401


prasek:
@eladb @iliapolo this is a follow-on to #378 and should get us to a clean L1 import for `crossplane-cdk` which I'll be working on next.

prasek:
was discussing with @eladb and @iliapolo -- think we'll make export module per apiGroup the default for all cases and skip the `--emit` option. I'll make those changes and verify this is working for other languages and should be good to go.

prasek:
@eladb verified the import works with python and Java. The exporter splits the `.` in the group name into different directories so thinking about adjusting the module names as follows:

Java: reverse order the group name for the `moduleName`to follow Java conventions, e.g.
- meta.pkg.crossplane.io -> imports/io/crossplane/pkg/meta

python: remove `.` from
the group name as the `moduleName`, e.g.
- meta.pkg.crossplane.io -> imports/metapkgcrossplaneio so we don't get a deep directory structure for python.

How does that sound?

eladb:
Java sounds good. For python maybe add underscores?

prasek:
sounds good, was just making that change to my comment ...

prasek:
@eladb although looking at https://www.python.org/dev/peps/pep-0008/#package-and-module-names:
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

and [PEP 420 -- Implicit Namespace Packages](https://www.python.org/dev/peps/pep-0420/) to see if using namespace packages would be better. 

Looks like it does work, so suggest reversing the group names for python as well - e.g. python imports would look like this:
```py
from constructs import Construct
from cdk8s import App, Chart
from imports import k8s
from imports.io.crossplane import pkg
from imports.io.crossplane.pkg import meta
```

eladb:
Yes looks great

prasek:
@eladb for consistency should reverse the group name for `import -l typescript` as well?

It current imports as:
```
├── apiextensions.crossplane.io.ts
├── k8s.d.ts
├── k8s.js
├── k8s.ts
├── meta.pkg.crossplane.io.ts
└── pkg.crossplane.io.ts
```

With group name reversed for typescript would be:
```
├── io.crossplane.apiextensions.ts
├── k8s.d.ts
├── k8s.js
├── k8s.ts
├── io.crossplane.pkg.meta.ts
└──io.crossplane.pkg.ts
```

eladb:
I am not sure about reversal at all (perhaps only in java). From a user's POV it feels more natural to refer to the group name as it is in the manifest, no?

prasek:
> I am not sure about reversal at all (perhaps only in java). From a user's POV it feels more natural to refer to the group name as it is in the manifest, no?

@eladb Fair point. I'm on the fence as well. With reversal it's nice to have the imported files grouped by project and for intellisense discovery of packages in a project, but agree it's not quite as natural coming from k8s and having to map back and forth.

I pushed a version that doesn't do any reversals to give it a try, but after trying it out would still recommend reversing for python and java.

As I was fixing up the examples I noticed that without reversal the python imports look like this:
```py
from imports.jenkins import io as jenkins
from imports.mattermost import com as mattermost
```

lmk on reversal for python and java, but otherwise should be good to go. thx.

eladb:
Okay let's go with reversal for python and java

prasek:
@eladb ready for review, thx. 
/cc @hasheddan

prasek:
Updated PR and commit title to reflect all crd imports emitting a single module per API group, not just crdsdev imports.

prasek:
@eladb I had already added some defensive code to consistently order the modules and CRDs, so we don't need a fix for https://github.com/crdsdev/doc/issues/110 as cdk8s now consistently orders CRDs in the output modules regardless of input manifest CRD ordering. As such have removed the extraneous doc.crds.dev/raw check that was tripping up the build and raised https://github.com/crdsdev/doc/issues/110 which can be handled separately.

Think we should be good to go now, thx.

eladb:
Awesome. I'll take a look tomorrow

eladb (approved):

PR #404, opened 2020-11-19
fix(plus-17): don't allow containers to be contructed from containers

abierbaum:
Add a check for the case where a container is attempted to be built from another container.

Any place that takes ContainerProps actually also accepts Containers. This happens because ContainerProps has all optional properties except image .  Since Container has an a public property image, it gets passed along but that is really the only thing that gets used. 

Problem is that if you pass in a container object, the nested container will not get all the settings, so I don't think this should be allowed.

note: this is a bit of an ugly hack to prevent this.  I was hoping to find a way to do this with the type system but didn't find a solid option without changing some public property names and breaking the API.

### DIFF
--- a/packages/cdk8s-plus-17/src/container.ts
+++ b/packages/cdk8s-plus-17/src/container.ts
@@ -285,6 +285,10 @@ export class Container {
   private readonly _startup?: Probe;
 
   constructor(props: ContainerProps) {
+    if (props instanceof Container) {
+      throw new Error('Attempted to construct a container from a Container object.');
+    }
+
     this.name = props.name ?? 'main';
     this.image = props.image;
     this.port = props.port;
--- a/packages/cdk8s-plus-17/test/container.test.ts
+++ b/packages/cdk8s-plus-17/test/container.test.ts
@@ -118,6 +118,17 @@ describe('Container', () => {
 
   });
 
+  test('Must use container props', () => {
+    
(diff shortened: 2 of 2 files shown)


packages/cdk8s-plus-17/src/container.ts
```diff
@@ -285,6 +285,10 @@ export class Container {
   private readonly _startup?: Probe;
 
   constructor(props: ContainerProps) {
+    if (props instanceof Container) {
```
eladb, line 288: How can this happen if the type of `props` is a ContainerProps?

eladb (commented):
What's the rationale?

packages/cdk8s-plus-17/src/container.ts
abierbaum: The test suite below shows an example.  The issue is that a Container object is considered a valid ContainerProps interface.  The reason is that anything is a valid ContainerProps that has an `image: string` property.  (see: https://github.com/awslabs/cdk8s/blob/aa466d46757e1e856c7c61710abef8424f53b9ef/packages/cdk8s-plus-17/src/container.ts#L149)

eladb (approved):

abierbaum:
@eladb I should say, I was really hoping to find a better way to do this.  I would expect the type system to be able to protect at compile time against this type of usage, but I couldn't find a good way to do it.

PR #415, opened 2020-11-24
fix(lib): Prevent substring of undefined error

dave-b-code:
Currently the calcHash method can be run on ApiObjects without an addr property, resulting in the cdk8s synth command failing due to a substring of an undefined variable. This PR prevents the cdk8s failing, and provides an empty string as a default return. Thanks.

### DIFF
--- a/packages/cdk8s/src/names.ts
+++ b/packages/cdk8s/src/names.ts
@@ -187,7 +187,11 @@ function calcHash(node: Node, maxLen: number) {
     return hash.digest('hex').slice(0, maxLen);
   }
 
-  return node.addr.substring(0, HASH_LEN);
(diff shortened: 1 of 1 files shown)


eladb (changes requested):
Why is this needed. If your project uses the required version of `constructs` then `node.addr` should be defined.

dave-b-code:
Hi, you are correct and I had not updated constructs when running the npm run upgrade command from an older version of cdk8s. I will close this PR as it is not necessary with the listed peer dependencies installed. Given that constructs is a required peer dependency, would it be possible to add it to the npm upgrade command to prevent this issue happening for other users upgrading using the built in script? Thanks.

PR #416, opened 2020-11-25
feat(cli): import-cr (WIP)

prasek:
Introduces a `cdk8s import-cr` command
- imports sample yaml CRs from a file or directory (recursive)
- generates L1 resource builders for testing L2 constructs or general use

example: L2 test [configuration.test.ts](https://github.com/prasek/crossplane-cdk/blob/cross-pkg/test/configuration.test.ts)

1. `cdk8s import -o src/crossplane/imports github:crossplane/crossplane@0.14.0`
2. `cdk8s import-cr -o test/crs test/fixtures`
3. write an L1 test with `expect(crL1).toMatchSnapshot()` with crL1 from the `cdk8s import-cr` output
4. write an L2 construct starting with an imported sample L1 CR to wrap and refactoring the L2 from there with TDD
5. add an L2 test with `expect(crL2).toStrictEqual(crL1)`

Notes:
- for (5) above would be nice to have a `Testing` method to output normalized YAML with consistent prop ordering instead of comparing via an intermediary `crL1`, e.g.:
```ts
const y1 = Testing.normalizedYaml(yamlL1);
const y2 = Testing.normalizedYaml(Testing.synth(crL2));
expect(y1).toStrictEqual(y2)
```

TODO:
- resolve typed enums
- resolve import paths
- validate original source yaml vs. generated yaml


prasek:
was discussing with @eladb and @iliapolo -- it may make more sense for this to go into a crossplane-specific importer but think there's value in importing existing CR yamls as a test scaffold for L2 types that you can use for TDD and then refactor based on your L2 simplifications, but also @eladb thought it would be useful to do a normalized yaml compare as part of `Testing` which would replace the need for this for simple L1 snapshot testing.

eladb:
@prasek I am closing for now. Ok?

PR #417, opened 2020-11-25
fix: which command missing for windows

campionfellin:
Closes https://github.com/awslabs/cdk8s/issues/414

Looked at these for some more info:

https://superuser.com/questions/207707/what-is-windows-equivalent-of-the-which-command-in-unix-is-there-an-equivale
https://stackoverflow.com/questions/8683895/how-do-i-determine-the-current-operating-system-with-node-js

---

It brings up the larger question though - should the `build` workflow run on Windows? Will the JSII/superchain docker image work on Windows?

---

### DIFF
--- a/packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js
+++ b/packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js
@@ -1,19 +1,19 @@
 const { execSync } = require('child_process');
 const { chmodSync } = require('fs');
 const { readFileSync } = require('fs');
-import { whichShim } from '../../src/util';
+const { platform } = require('os');
 
 const cli = require.resolve('../../bin/cdk8s');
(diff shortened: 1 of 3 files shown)


packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js
```diff
@@ -1,20 +1,21 @@
 const { execSync } = require('child_process');
 const { chmodSync } = require('fs');
 const { readFileSync } = require('fs');
+import { whichShim } from '../../src/util';
```
eladb, line 4: Does that import even work?

campionfellin: Nope! Fixed it, good callout

eladb (dismissed):

PR #421, opened 2020-11-26
fix: Wrong name for classes in python

tsahiduek:
When tried using python (havn't checked other languages) -->
`k8s.KubeService` should be `k8s.Service`
`k8s.KubeDeployment` should be `k8s.Deployment`


iliapolo:
@tsahiduek Thanks for this, but we actually recently [changed](https://github.com/awslabs/cdk8s/pull/370) the import logic, which now generates the `Kube` prefix.  Make sure you re-import the `k8s` file with the latest version of the CLI.

PR #424, opened 2020-11-26
fix(plus-17): adds externalName to service props

pusherman:

- Adds externalName to service props
- Validates externalName is set when using an EXTERNAL_NAME type
- Adds externalIPs test to get back to 100% coverage in the service tests


packages/cdk8s-plus-17/src/service.ts
```diff
@@ -232,6 +244,10 @@ export class Service extends Resource {
       throw new Error('A service must be configured with a port');
     }
 
+    if (this.type === ServiceType.EXTERNAL_NAME && this.externalName === undefined) {
+      throw new Error('A service with type EXTERNAL_NAME requires an externalName prop');
```
eladb, line 248: Should we also check the opposite? I.e if an externalName is specified then the type must either be undefined or SERVICE_NAME?

eladb (changes requested):

packages/cdk8s-plus-17/src/service.ts
pusherman: Added the inverse check in as well

packages/cdk8s-plus-17/src/service.ts
```diff
@@ -232,6 +244,14 @@ export class Service extends Resource {
...
+    }
+
+    if (this.externalName !== undefined && this.type !== ServiceType.EXTERNAL_NAME) {
```
eladb, line 251: I'd say that if `type` is undefined but `externalName` is defined then we can safely imply that type is `EXTERNAL_NAME`. Only if type is explicitly set to something else then we want to fail.

eladb: This is also what I proposed in the previous comment

packages/cdk8s-plus-17/test/service.test.ts
```diff
@@ -138,3 +138,48 @@ test('Cannot add a deployment if a selector is already defined for this service'
...
+
+  service.serve(5432);
+
+  expect(() => Testing.synth(chart)).toThrowError(
+    'A service with type EXTERNAL_NAME requires an externalName prop',
+  );
+
+});
+
+test('Must be configured with a type of EXTERNAL_NAME if externalName if given', () => {
```
eladb, line 171: A test where type is explicitly set to something else

eladb (changes requested):

packages/cdk8s-plus-17/src/service.ts
pusherman: @eladb - how do you feel about checking to see if externalName is set in the props in the constructor.  If it is, we force the type to EXTERNAL_NAME.  Since the only way set the type, we should be safe without the check in `_toKube`

Something like
```
    if (props.externalName !== undefined) {
      this.type = ServiceType.EXTERNAL_NAME;
    } else {
      this.type = props.type ?? ServiceType.CLUSTER_IP;
    }
```

abierbaum:
@eladb @pusherman Anything still pending on this PR?  I would love to start using it in production instead of my other hack.

eladb (approved):
Sorry slipped my radar. Thanks for pinging

PR #425, opened 2020-11-28
feat(plus-17): Add type option for secrets in kplus.

abierbaum:
Add the type option for kplus.Secret.  I wanted to use this to make a TLS cert secret with the normal `kubernetes.io/tls` type.


packages/cdk8s-plus-17/src/secret.ts
```diff
@@ -11,6 +11,14 @@ export interface SecretProps extends ResourceProps {
...
    */
   readonly stringData?: { [key: string]: string };
+
+  /**
+   * Optional type associated with the secret.  Used to facilitate programmatic
+   * handling of secret data by various controllers.
+   *
+   * @default undefined - Don't set a type.
+   */
+  readonly type?: string;
```
eladb, line 21: Does it make sense to make this an enum? Are there a specific list of supported types?

abierbaum: There isn't an exhaustive list, but there are some common built in types.  I think a reasonable option would be having type exposed like this but then adding some static helpers to build up the common settings.

so for example:

```
public static forTls(scope: Construct, id: string, crt: string, key: string): Secret {
   return new Secret(scope, id, {
   type: 'kubernetes.io/tls',
      stringData: {
       'tls.crt': crt,
       'tls.key': key,
     }
   })
}
```

abierbaum:
@eladb What do you think about getting this PR through?  Do you want the helpers first or what are the next steps?

eladb (approved):

PR #427, opened 2020-11-29
fix(cli): importing local files is broken on windows

dave-b-code:
Previously protocol was used to detect if file is local or on a remote url, however windows absolute file paths start with C: or D: etc. so are incorrectly detected as urls leading to error on line 46. fs.lstatSync(url).isFile() seemed to be the best option from what I can tell, but fs.existsSync() or others may be better. Thanks in advance for the library and any feedback on this PR. .


packages/cdk8s/src/_loadurl.js
```diff
@@ -17,7 +17,16 @@ if (!url) {
...
+let isLocalFile = false;
+try {
+  if(fs.lstatSync(url).isFile()){
+    isLocalFile = true;
+  }
+} catch (error) {
+  isLocalFile = false;
+}
+
+if (isLocalFile) {
```
iliapolo, line 29: If it's a local file, shouldn't we just pass `url` to `fs.createReadStream` below? and in that case we can remove the `if purl.pathname` check.

will `pathname` work the same?

The flow is a little confusing to me, can we first check if it's a local file, and if so just call `createReadStream`, otherwise, parse the url, validate we have a protocol, and call the `http` module.

WDYT?

iliapolo (changes requested):

packages/cdk8s/src/_loadurl.js
dave-b-code: Hi, thanks for the feedback, the initial code was indeed pretty verbose so I've trimmed it and implemented the changes you requested. Please let me know if there's anything else I can do further to improve it.

packages/cdk8s/src/_loadurl.js
```diff
@@ -17,13 +17,11 @@ if (!url) {
...
-  if (!purl.pathname) {
-    throw new Error(`unable to parse pathname from file url: ${url}`);
+try {
+  if (fs.lstatSync(url).isFile()) {
+    fs.createReadStream(url).pipe(process.stdout);
   }
-
-  fs.createReadStream(purl.pathname).pipe(process.stdout);
-} else {
+} catch (err) {
```
iliapolo, line 24: I don't think we should drop the `if (!purl.protocol)` check. Just move it here along with the call to `parse`, which isn't needed if its a local file.

packages/cdk8s/src/_loadurl.js
```diff
@@ -15,13 +15,13 @@ if (!url) {
...
 }
 
+const purl = parse(url);
+
 try {
   if (fs.lstatSync(url).isFile()) {
     fs.createReadStream(url).pipe(process.stdout);
   }
 } catch (err) {
-  const purl = parse(url);
```
iliapolo: Why did you end up undoing this?

dave-b-code: Because the function getHttpClient() uses purl so it needs to be scoped appropriately. Could alternatively turn getHttpClient into a pure function and move parse line, I'll add a commit with that just in case.

iliapolo: 👍

iliapolo (approved):

PR #436, opened 2020-12-05
chore: projen mono repo configuration and auto dependency upgrades

iliapolo:


tools/pack.sh
```diff
@@ -4,10 +4,7 @@ repo_root=$(cd $(dirname $0)/.. && pwd)
 cd ${repo_root}
 
 echo "Cleaning dist directories..."
-rm -rf dist/
-rm -rf packages/cdk8s/dist
-rm -rf packages/cdk8s-plus-17/dist
-rm -rf packages/cdk8s-cli/dist
+rm -rf **/dist
```
eladb, line 7: Convert `pack.sh` to a projen Task and derive these `rm -rf` execs from the project structure.

What if a `node_module` includes a `dist` directory?

eladb (changes requested):
Feels like you could have saved 1000 hacks if you'd just implement `NodeMonoRepo` inside `projen` directly.

PR #440, opened 2020-12-10
feat(plus-17): add ResourceRequirements to Container

foriequal0:
HPA requires containers to set resource requests and limits. I've added a Cpu unit type and a way to define resource requirements to Container.

Closes: https://github.com/awslabs/cdk8s/issues/438

### DIFF
--- a/packages/cdk8s-plus-17/src/container.ts
+++ b/packages/cdk8s-plus-17/src/container.ts
@@ -1,3 +1,4 @@
+import * as cdk8s from 'cdk8s';
 import { IConfigMap } from './config-map';
 import * as k8s from './imports/k8s';
 import { Probe } from './probe';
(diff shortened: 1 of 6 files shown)


packages/cdk8s-plus-17/src/container.ts
```diff
@@ -508,6 +518,70 @@ export enum MountPropagation {
   BIDIRECTIONAL = 'Bidirectional',
 }
 
+/**
+ * Properties for creating ResourceRequirements
+ */
+export interface ResourceRequirementsProps {
+  readonly cpu?: IResourceRequirement<cdk8s.Cpu>;
```
iliapolo, line 525: We avoid using generics in public interfaces because they don't translate well to `jsii`. Create separate types.

iliapolo (changes requested):

foriequal0:
Sorry I don't have much time for this PR for a while. I should close this PR for now.

PR #441, opened 2020-12-11
feat: ci for all platforms

campionfellin:
First part of https://github.com/awslabs/cdk8s/issues/422 to enable CI for windows.

The tests will fail, but then we can start working on Windows support from there.


test/test-imports/common.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
eladb, line 1: This script is intended to be `source`d so shebang is not required (and actually misleading)

test/test-imports/python/test.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
eladb, line 1: Same

test/extract-dist.sh
```diff
@@ -23,6 +23,4 @@ npm init -y > /dev/null
 npm install ${CDK8S_DIST}/js/*.tgz
 export PATH=${staging}/node_modules/.bin:$PATH
 
-# restore working directory
```
eladb, line 26: Not sure what's up here. run-against-dist is useful for locally running integ tests against the dist folder.

eladb (changes requested):

test/extract-dist.sh
campionfellin: The reason I split it up here was because the `PATH` was not getting carried over from here to the `test-all.sh` script. One way to get the path working was to add it to `$GITHUB_PATH`, but anything added there cannot be used in the same step. So then I split the extract and test into two steps.

packages/cdk8s-cli/templates/python-app/.hooks.sscaff.js
```diff
@@ -47,8 +45,8 @@ exports.post = options => {
 
   chmodSync('main.py', '700');
 
-  execSync(`${cli} import k8s -l python`);
-  execSync(`pipenv run ./main.py`);
+  execSync(`cdk8s import k8s -l python`);
+  execSync(`pipenv run "./main.py"`);
```
campionfellin, line 49: Getting nearly there. Failing here with:

```
'.' is not recognized as an internal or external command,
```

Likely some windows path magic to fix

campionfellin:
After some more digging in a separate branch I'm using, it seems like the only tests that fail are these python tests. Will continue hammering away, nearly there!

campionfellin:
@iliapolo this should be ready to go!

test/test-imports/typescript/test.sh
```diff
@@ -1,4 +1,3 @@
-#!/bin/sh
```
eladb, line 1: why?

packages/cdk8s-cli/templates/java-app/.hooks.sscaff.js
```diff
@@ -35,7 +33,7 @@ exports.post = options => {
   }
 
   execSync(`mvn install`);
-  execSync(`${cli} import k8s -l java`);
+  execSync(`cdk8s import k8s -l java`);
```
eladb, line 36: This will search the path instead of use the _current_ CLI. What's the rationale?

.github/workflows/build.yml
```diff
@@ -25,6 +25,40 @@ jobs:
...
+          path: dist
+
+      - name: extract dist
+        run: yarn extract
+        shell: bash
+
       - name: integration tests
         run: yarn integ
-
+        shell: bash
```
eladb, line 64: what does that do?

test/test-imports/typescript/test.sh
campionfellin: Based on some of your previous comments, I thought I'd remove it. It's a `.sh` file and the shebang seems redundant here. Open to putting it back though.

packages/cdk8s-cli/templates/java-app/.hooks.sscaff.js
campionfellin: On Windows it has trouble resolving `../../bin/cdk8s` so I switched to `cdk8s` from the path. These templates are used during the `cdk8s init ...` command and I'm making the assumption most users have it in their path rather than like `/usr/local/bin/cdk8s init ...`

.github/workflows/build.yml
campionfellin: The default shell for non-Windows is `bash` and for Windows is `powershell`. Explicitly setting this so that the Windows build uses bash. Here is the `bash` info for Github Actions:

```
The default shell on non-Windows platforms with a fallback to sh.
When specifying a bash shell on Windows, the bash shell included with Git for Windows is used.
```


Docs: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

test/test-imports/typescript/test.sh
campionfellin: After putting it back, seems to cause this error on ubuntu:

```
python/test.sh: 6: python/test.sh: source: not found
```

Going to revert back to having it removed.

campionfellin:
Need to resolve this one in `projen` first - https://github.com/projen/projen/issues/498

eladb:
Ping

campionfellin:
Hey @eladb seems like recent versions of `projen` don't work on Windows. A few options we can try to unblock this:

* Downgrade `projen` back to v0.13.9
* Bump node up to v15 (just for Windows)
* Install `projen` globally on the github runner

Any other suggestions or ideas?

eladb:
Any specific issue? We can fix...

campionfellin:
Yeah, seems like `xmlbuilder2` takes a dependency on `@types/node 14.6.2`. I've forked it to take a `*` dependency, I think that might help.

Do you mind if we switch `projen`'s dependency to @campionfellin/xmlbuilder2 ? https://www.npmjs.com/package/@campionfellin/xmlbuilder2

eladb:
@campionfellin closing for now. reopen when ready to continue.

PR #442, opened 2020-12-11
feat(plus-17): add named port, port objects, exposing multiple ports

foriequal0:
Container can expose multiple ports, also it can name its ports.
Container and Deployment would validate the uniqueness of ports and port names. Kubernetes says they must be unique, but doesn't validate them, so ports can be silently shadowed, and can expose wrong port.
Probe and Services can expose by a port number, a port name, and a port object.
Container owns ContainerPort objects (Deployments owns them indirectly) and validate ownership by object reference even if the object seems to somewhat valid to prevent to expose wrong port with same properties.

Closes: https://github.com/awslabs/cdk8s/issues/439

### DIFF
--- a/packages/cdk8s-plus-17/src/container.ts
+++ b/packages/cdk8s-plus-17/src/container.ts
@@ -395,25 +395,36 @@ export class Container {
    * @param targetPort - Number or name of exposed port
    * @param throwOnNotfound
    */
-  public lookupPort(targetPort: number | string, throwOnNotfound: true): ContainerPort
-  public lookupPort(targetPort: number | string, throwOnNotfound?: boolean): ContainerPort | undefined
-  public lookupPort(targetPort: number | string, throwOnNotfound: boolean = false): ContainerPort | undefined {
+  public lookupPort(targetPort: number | string | ContainerPort, throwOnNotfound: true): ContainerPort
+  public lookupPort(targetPort: number | string | ContainerPort, throwOnNotfound?: boolean): ContainerPort | undefined
+  public lookupPort(targetPort: number | string | ContainerPort, throwOnNotfound: boolean = false): ContainerPort | undefined {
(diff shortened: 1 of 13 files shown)


packages/cdk8s-plus-17/src/container.ts
```diff
@@ -248,9 +257,9 @@ export interface ContainerProps {
 export class Container {
 
   /**
-   * The port this container exposes.
+   * List of ports this container exposes.
    */
-  public readonly port?: number;
+  public readonly ports: ContainerPort[] = [];
```
foriequal0, line 262: This change would be the breaking change. Should I create ```public get port() { return this.ports[0]?.port; }``` not to bring the breaking change?

iliapolo (commented):
@foriequal0 Thanks for this. It's going to take a little while until I can properly go over this. 

In the meantime, can you please refrain from using type unions in public API signatures? It translates poorly to other languages (basically as a generic `Object`) and we only do it as a last resort. 

We can discuss specific alternatives if you find yourself a little stuck.

foriequal0:
I got it. I think that would be introducing `interface IContainerPortResolver { resolve(IContainerPort): ContainerPort }` for `Container` and `Deployment`, with `NamedContainerPort: IContainerPort`, `NumericContainerPort: IContainerPort`, `ContainerPort: IContainerPort`, and `ContainerPort.fromName(string): NamedContainerPort`, `ContainerPort.fromPort(number): NumericContainerPort`

iliapolo:
@foriequal0 The comment is a little hard to read but yeah use static `from` methods to simulate union types.

foriequal0:
Thanks. Can I ask you to prioritize https://github.com/awslabs/cdk8s/pull/440 more than this?

foriequal0:
Sorry I don't have much time for this PR for a while. I should close this PR for now.

PR #446, opened 2020-12-13
feat(plus-17): restrict CIDR IP addresses for a LoadBalancer

### LINKED ISSUE #435: restrict CIDR IP addresses for a LoadBalancer type not supported
I would like to restrict my service to whitelisted cdir IP ranges. This is already supported. See the `loadBalancerSourceRanges` on the following page:
https://aws.amazon.com/premiumsupport/knowledge-center/eks-cidr-ip-address-loadbalancer/

Right now there is no way to specify it using the current object configurations. Thus forcing me to use an escape hatch.

### Use Case:
Increase security by whitelisting certain IPs.

### Proposed Solution:
I'm not sure which object you should put this on, but I should be able to configure it on one of the "high level" apis  (cdk8s-plus or cdk8s-plus-17) without having to apply an escape hatch.

### Other:

* [ ] :wave: I may be able to implement this feature request

---

saudkhanzada:
Fixes #435


iliapolo:
@saudkhanzada Sorry this slipped my radar. It looks good, just need to resolve conflict and i will approve.

iliapolo (approved):

PR #447, opened 2020-12-14
feat(cli): enable using imported resources as raw manifests

prasek:
Splits out a static `MyCrd.manifest(props: MyCrdProps)` from imported CRD class constructors,
which include `kind` and `apiVersion` with a strongly-typed props object. This can be used to
implement inline resource templates that include imported resources.

Example: https://github.com/crossplane-contrib/crossplane-cdk/pull/7.

```ts
composition.addResource(db.RdsInstance.manifest({
  spec: {
    forProvider: {
      region: 'us-west-2',
      dbSubnetGroupNameSelector: {
        matchControllerRef: true,
      },
      dbInstanceClass: 'db.t2.small',
      masterUsername: 'masteruser',
      engine: 'postgres',
      engineVersion: '9.6',
      skipFinalSnapshotBeforeDeletion: true,
      publiclyAccessible: false,
    },
    writeConnectionSecretToRef: {
      namespace: 'crossplane-system',
      name: 'default-db-conn'
    },
    deletionPolicy: db.RdsInstanceSpecDeletionPolicy.DELETE,
  }
}));
```

Output:

```yaml
apiVersion: v1
kind: RdsInstance
spec:
  forProvider:
    region: us-west-2
    dbSubnetGroupNameSelector:
      matchControllerRef: true
...
```

Signed-off-by: Phil Prasek <prasek@gmail.com>


packages/cdk8s-cli/src/import/codegen.ts
```diff
@@ -87,24 +91,33 @@ export function generateConstruct(typegen: TypeGenerator, def: ApiObjectDefiniti
...
 
-    function emitInitializerSuper() {
+    function emitPropsWithGVK() {
+
+      code.line('/**');
+      code.line(` * Adds "${def.fqn}" kind and apiVersion to props`);
+      code.line(' * @param props initialization props');
+      code.line(' */');
+
+      code.openBlock(`public static propsWithGVK(props: ${propsTypeName}${defaultProps}): any`);
```
eladb, line 111: Instead of a method which merges GVK into an arbitrary props objects, why not just emit something like this:

```ts
export class KubePod extends ... {
  public static readonly GVK: GroupVersionKind = {
    apiVersion: 'v1',
    kind: 'Pod'
  };
}
```

Then, the caller can merge that into anything they want.

eladb (changes requested):

packages/cdk8s-cli/src/import/codegen.ts
```diff
@@ -76,35 +79,65 @@ export function generateConstruct(typegen: TypeGenerator, def: ApiObjectDefiniti
       code.line(' */');
       code.openBlock(`export class ${constructName} extends ApiObject`);
 
+      emitGVK();
```
prasek, line 82: @eladb I added both here so we can get cleaner usage with `propsWithGVK()`

packages/cdk8s-cli/src/import/codegen.ts
prasek: I added both here so we can get cleaner usage with propsWithGVK()

iliapolo (approved):

PR #459, opened 2020-12-16
feat: add YAML parse and stringify

### LINKED ISSUE #419: Yaml.stringify()
I would love to have some YAML utilities built-in to cdk8s.

### Use Case:

While working with k8s, we are working with YAML in one way or another. Having some YAML utilities would enable users to do things like this one, without having to install external deps:
```typescript
    const datasources = new ConfigMap(this, 'grafana-datasources', {
      data: {
        "loki.yaml": yaml.safeDump({ 
          apiVersion: 1, 
          datasources: [{
            name: 'Loki', 
            type: 'loki', 
            access: 'proxy', 
            url: `${loki.releaseName}:3100`, 
            jsonData: { maxLines: 1000 } 
          }]
        })
      }
    });
```

### Other:

This was proposed by @eladb on Slack.

* [ ] :wave: I may be able to implement this feature request -> I don't have enough bandwidth right now :(

---

dgzlopes:
Fixes #419 

This PR exposes two YAML methods: parse and stringify.

I didn't add tests b/c it's just a wrapper, but if needed, I can add some.

### DIFF
--- a/packages/cdk8s/src/yaml.ts
+++ b/packages/cdk8s/src/yaml.ts
@@ -85,7 +85,7 @@ export class Yaml {
    * @param str The YAML String.
    * @returns A value that matches the type of the root value of the parsed YAML document.
    */
-  public static parse(str: string): any{
+  public static parse(str: string): any {
     return YAML.parse(str);
   }
 
--- a/packages/cdk8s/src/yaml.ts
+++ b/packages/cdk8s/src/yaml.ts
@@ -76,7 +76,7 @@ export class Yaml {
    * @param value The set of objects.
    * @returns A String with YAML formatting.
    */
-  public static stringify(value: any): string{
+  public static stringify(value: any): string {
     return YAML.stringify(value);
   }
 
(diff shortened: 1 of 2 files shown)


packages/cdk8s/src/yaml.ts
```diff
@@ -71,6 +71,24 @@ export class Yaml {
     return result;
   }
 
+  /**
+   * Converts a set of objects to a YAML string.
+   * @param value The set of objects.
+   * @returns A String with YAML formatting.
+   */
+  public static stringify(value: any): string{
```
eladb, line 79:
```suggestion
  public static stringify(value: any): string {
```

packages/cdk8s/src/yaml.ts
```diff
@@ -71,6 +71,24 @@ export class Yaml {
...
+    return YAML.stringify(value);
+  }
+
+  /**
+   * Converts a String with YAML formatting to a set of objects.
+   * @param str The YAML String.
+   * @returns A value that matches the type of the root value of the parsed YAML document.
+   */
+  public static parse(str: string): any{
```
eladb, line 88:
```suggestion
  public static parse(str: string): any {
```

eladb (changes requested):
Please add unit tests

PR #461, opened 2020-12-16
feat(plus-17): Add support for externalName on Service

abierbaum:
This add support for Services that are of type EXTERNAL_NAME.

I pushed this up before I saw the other PR https://github.com/awslabs/cdk8s/pull/424 from @pusherman.

There may be some ideas here to merge into that one or vice versa.  My goal is simply to get support for external name through.

### DIFF
--- a/packages/cdk8s-plus-17/src/service.ts
+++ b/packages/cdk8s-plus-17/src/service.ts
@@ -35,6 +35,15 @@ export interface ServiceProps extends ResourceProps {
    */
   readonly externalIPs?: string[];
 
+   /**
+   * The external reference that kubedns or equivalent will return as a CNAME record
+   * for this service. No proxying will be involved. Must be a valid RFC-1123 hostname
+   * (https://tools.ietf.org/html/rfc1123) and requires Type to be ExternalName.
+   *
+   * @default - No external name.
+   */
+  readonly externalName?: string;
+
   /**
    * Determines how the Service is exposed.
    *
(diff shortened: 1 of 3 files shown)


packages/cdk8s-plus-17/src/service.ts
```diff
@@ -244,12 +259,16 @@ export class Service extends Resource {
       });
     }
 
-    return {
+    return this.type !== ServiceType.EXTERNAL_NAME ? {
```
abierbaum, line 262: As compared to https://github.com/awslabs/cdk8s/pull/424, I tried to ensure that the spec returned only has the details really needed.

pusherman: I actually like this approach better :)

iliapolo: @abierbaum I see that @pusherman PR contains a few additional stuff i'd like to also merge, so lets to merge that one instead? 

@pusherman are you waiting for another review on that PR?

abierbaum: @iliapolo Sounds good to me.  Looks like @pusherman liked this one piece that seemed like the only improvement I had.  If he wants to role that into his PR that is great but I am also fine without it.  Just need something to use. :)

iliapolo: Yeap, lets give @@pusherman a day or two to respond :)

pusherman: just updated my PR to include the code to only return the needed props needed for an ExternalService.  It's ready to review whenever it passes the build/test checks.  Thanks @iliapolo and @abierbaum!

PR #465, opened 2020-12-16
chore(deps): Upgrade transitive dependencies

iliapolo:
The current workflow doesn't upgrade transitive dependencies because it only runs `yarn install` and not `yarn upgrade`.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -18,8 +18,11 @@ jobs:
         # These need to be ignored from the `ncu` runs since local dependencies
         # need to remain on 0.0.0
         run: |-
+
+          export PATH=./node_modules/.bin:${PATH}
+
           echo -n "::set-output name=list::"
-          node -p "$(./node_modules/.bin/lerna ls --all --json 2>/dev/null).map(item => item.name).join(',')"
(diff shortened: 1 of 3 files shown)


.github/workflows/dependencies.yml
```diff
@@ -13,19 +13,24 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 10.17.0
-      - name: Install Tools
+      - name: Install
         run: |-
-          npm -g install lerna npm-check-updates@^9.0.0
```
iliapolo, line 18: Moved this as a dependency of the root `package.json` so that its also easily available locally.

.github/workflows/dependencies.yml
```diff
@@ -46,8 +51,7 @@ jobs:
           lerna exec --parallel ncu -- --upgrade --reject='constructs,projen,typescript,${{ steps.list-packages.outputs.list }}' --target=minor
 
       - name: Upgrade lock file
-        # This will create a brand new `yarn.lock` file (this is more efficient than `yarn install && yarn upgrade`)
-        run: yarn install --force
```
iliapolo, line 50: `install --force` didn't update any packages that hadn't explicitly changed in one the `package.json` files, i.e transitive dependencies. We need those upgrades as well.

eladb (approved):

PR #489, opened 2020-12-25
fix(plus-17): multiple mounts per volume result in duplicate volumes for pod spec

iliapolo:
Allow multiple container mounts to use the same volume.

This PR also introduces a restriction for adding two different volumes with the same name to a pod.

Fixes https://github.com/awslabs/cdk8s/issues/483

### DIFF
--- a/packages/cdk8s-plus-17/src/pod.ts
+++ b/packages/cdk8s-plus-17/src/pod.ts
@@ -112,9 +112,7 @@ export class PodSpec implements IPodSpec {
       throw new Error('PodSpec must have at least 1 container');
     }
 
-    const volumes: Set<Volume> = new Set();
-
-    const k8sVolumes: k8s.Volume[] = [];
+    const volumes: Map<string, k8s.Volume> = new Map();
     const containers: k8s.Container[] = [];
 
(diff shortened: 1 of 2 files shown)


packages/cdk8s-plus-17/src/pod.ts
```diff
@@ -112,29 +112,33 @@ export class PodSpec implements IPodSpec {
...
+        }
       }
 
       containers.push(container._toKube());
     }
 
     for (const volume of this._volumes) {
-      volumes.push(volume._toKube());
+      if (!volumes.has(volume.name)) {
+        volumes.set(volume.name, volume._toKube());
```
eladb, line 133: Should we ensure this is actually the same volume (toKube returns the same thing or the Volume instance is the same)? If they don't match I'd fail

eladb (changes requested):

packages/cdk8s-plus-17/src/pod.ts
iliapolo: Good idea

eladb (approved):

PR #498, opened 2020-12-31
fix(lib): Update imports so they will work with TS esModuleInterop

abierbaum:
The imports as previously defined would run into an error when imported by projects that use the current tsconfig.json default of true for compilerOptions esModuleInterop.

```
Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.
```

This changes the code to use a require and thus allow usage with projects using interop.

I suggest considering enabling esModuleInterop for the project as a whole as this is the direction Typescript has gone with their defaults.  Although I confess this is not a place I have a strong understanding.  I am going to include a few links that I ran across though in my investigation of why this wasn't working for me.  This could help others decide if it would make sense.  In any case, changing these two imports allows the code to work for people using interop.

References:  (sorry to reference stackoverflow, but it was the most clear summaries I could find)

   * TS 2.7 Change: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#support-for-import-d-from-cjs-from-commonjs-modules-with---esmoduleinterop
   * Example handling the same issue: https://stackoverflow.com/questions/49256040/a-namespace-style-import-cannot-be-called-or-constructed-and-will-cause-a-failu
   * Good answer with details: https://stackoverflow.com/a/56348146/205103

### DIFF
--- a/packages/cdk8s-plus-17/src/config-map.ts
+++ b/packages/cdk8s-plus-17/src/config-map.ts
@@ -2,7 +2,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as cdk8s from 'cdk8s';
 import { Construct } from 'constructs';
-import * as minimatch from 'minimatch';
+import minimatch = require('minimatch');
 import { ResourceProps, Resource, IResource } from './base';
 import * as k8s from './imports/k8s';
 import { undefinedIfEmpty } from './utils';
--- a/packages/cdk8s/src/api-object.ts
+++ b/packages/cdk8s/src/api-object.ts
@@ -1,5 +1,5 @@
 import { Construct, IConstruct, Node } from 'constructs';
-import * as stringify from 'json-stable-stringify';
+import stringify = require('json-stable-stringify');
 import { resolve } from './_resolve';
 import { sanitizeValue } from './_util';
 import { Chart } from './chart';


abierbaum:
Note: lint is failing because lint is set to not allow require() calls.  That is the only way I have found to allow interop to work and not break code for people without interop.  Hopefully someone else knows a better way.

abierbaum:
@iliapolo Is this something that could be supported by cdk8s?  It looks like the use of esModuleInterop has been the Microsoft recommendation for Typescript for quite a while.

iliapolo:
@abierbaum Sorry this slipped my radar. 

Honestly i'm not fully versed in all these compiler options, so I don't yet have a definitive answer for you. I need to dive deeper here. 

I'm trying to understand the problem though, are you saying that projects who have `esModuleInterop: true` are failing to import `cdk8s` because of the two imports you point to?

abierbaum:
> I'm trying to understand the problem though, are you saying that projects who have `esModuleInterop: true` are failing to import `cdk8s` because of the two imports you point to?

@iliapolo Yes, that is what I ran into anyway.  I looked into for about an hour and the best I could find was to change the two imports to make them be able to work regardless of which way the flag is set.

iliapolo:
@abierbaum Ok, let me take a deeper look next week and we'll see how to approach this.

abierbaum:
@iliapolo Sounds great.  I appreciate it.

abierbaum:
@iliapolo Did you have any time to look into this?  It is the only customization I still have left that requires me to us a custom branch of cdk8s.  I would love to get back to just using mainline.

iliapolo:
@abierbaum Sorry I haven't delved into this yet. Thanks for checking in, lets invite @RomainMuller to weigh in, I feel he might have some insights :).

@RomainMuller what do you think about dropping the `@typescript-eslint/no-require-imports` from ESLint? Is there another way to support `esModuleInterop`?

abierbaum:
Has anyone else found a workaround to this?  I am still maintaining a custom fork of cdk8s just to carry these two changes and I would like to get rid of that overhead.

iliapolo:
@abierbaum Can you clarify exactly where you encounter the problem? I've tried creating a app that uses `cdk8s-plus` with `esModuleInterop` enabled and was able to import `ConfigMap` and properly synth. 

Also, can you go ahead and add eslint exclusions to those two lines and make the build pass? (note there are also merge conflicts now) - I'm inclined to merge to this PR so it would be good to have it ready. 

Thanks!

abierbaum:
@iliapolo I have created a demonstration that shows the problem: https://github.com/abierbaum/cdk8s_esmoduleinterop_bug

And now that I have, it is strange.  The issue does not occur when using the cdk8s packages directly.  It only crops up when using a checkout of the cdk8s github repo and using the tools/link-all.sh command to have the project make use of the checkout.

To reproduce it, checkout the repository above and then:

```
> git submodule init
> git submodule update
> yarn install
> yarn link_deps
> yarn compile
```

I get:

```
 yarn compile
yarn run v1.22.5
$ tsc
deps/cdk8s/packages/cdk8s-plus-17/lib/config-map.ts:153:13 - error TS2349: This expression is not callable.
  Type 'typeof M' has no call signatures.

153         if (minimatch(file, pattern)) {
                ~~~~~~~~~

  deps/cdk8s/packages/cdk8s-plus-17/lib/config-map.ts:5:1
    5 import * as minimatch from 'minimatch';
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.

deps/cdk8s/packages/cdk8s/node_modules/yaml/dist/nodes/Pair.d.ts:14:9 - error TS2611: 'commentBefore' is defined as a property in class 'NodeBase', but is overridden here in 'Pair<K, V>' as an accessor.

14     get commentBefore(): string | null | undefined;
           ~~~~~~~~~~~~~

deps/cdk8s/packages/cdk8s/node_modules/yaml/dist/nodes/Pair.d.ts:16:9 - error TS2611: 'spaceBefore' is defined as a property in class 'NodeBase', but is overridden here in 'Pair<K, V>' as an accessor.

16     get spaceBefore(): boolean | undefined;
           ~~~~~~~~~~~


Found 3 errors.

error Command failed with exit code 1.
```

So now there are more than just the interop errors.  I fully admit that this now goes beyond my understanding of the typescript toolchain.  The reason we have a setup like this is because we had a few branches that were in PR still with cdk8s and needed a way for our production code to use the PRs.  I ended up setting up the repo with a submodule that pointed to a branch we have in github with our changes in it.  

There may be a smarter way to do this.  Although it was *very* useful to allow me to work on contributions / fixes to cdk8s while testing with our production code.

At this point I think I could just move over to the release version and probably be in a good spot.  Although I would be interested in understanding if there is a different way to make this work for the next time we need to track a branch with some of our changes while they are still in PR review or pre-release.

PR #499, opened 2020-12-31
chore(deps): Upgrade major version of jest

iliapolo:
To resolve the [security alert](https://github.com/awslabs/cdk8s/security/dependabot/yarn.lock/node-notifier/open), we need to upgrade the major version of jest because version `25` only works with `node-notifier` version `^6.0.0`.

This wasn't picked up by the automatic dependency upgrades because its a major version upgrade.


iliapolo:
Not ready yet

iliapolo:
ready now. @eladb

eladb (approved):

PR #510, opened 2021-01-18
feat(plus-17): add ingress sslRedirect method

### LINKED ISSUE #509: (cdk8s-plus ingress) support sslRedirect method.
`cdk8s-plus`  support `cdk8s-plus.IngressBackend.sslRedirect()` method.

### Use Case:

In `aws-load-balancer-controller` config, we need to add ssl-redirect rule in rules. 
https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/tasks/ssl_redirect/ 
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  namespace: default
  name: ingress
  annotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:us-west-2:xxxx:certificate/xxxxxx
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS":443}]'
    alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
spec:
  rules:
    - http:
        paths:
         - path: /*
           backend:
             serviceName: ssl-redirect
             servicePort: use-annotation
         - path: /users/*
           backend:
             serviceName: user-service
             servicePort: 80
         - path: /*
           backend:
             serviceName: default-service
             servicePort: 80
```
### Proposed Solution:

add `cdk8s-plus.IngressBackend.sslRedirect()` method to add  ssl-redirect  rule .
```ts
    const ACM = 'arn:aws:acm:us-east-1:0123456789:certificate/1002e2eb-3444-2222-1111-xxxxxxxxxx';
    const svc = new kplus.Service(this, 'svc', {
      ports: [{port: 3000}],
    });
    const ing = new kplus.Ingress(this, 'ing', {
      metadata: {
        name: 'k8sing',
        annotations: {
          'kubernetes.io/ingress.class': 'alb',
          'alb.ingress.kubernetes.io/load-balancer-attributes': 'idle_timeout.timeout_seconds=300',
          'alb.ingress.kubernetes.io/scheme': 'internet-facing',
          'alb.ingress.kubernetes.io/certificate-arn': ACM,
          'alb.ingress.kubernetes.io/listen-ports': '[{"HTTP": 80}, {"HTTPS":443}]',
          'alb.ingress.kubernetes.io/actions.ssl-redirect': '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}',
        },
      },
    });
    ing.addRule('/*',kplus.IngressBackend.sslRedirect());
    ing.addHostRule('xxx.example.com', '/*',kplus.IngressBackend.fromService(svc));
  }
```
### Other:

---

guan840912:
Close #509


iliapolo:
@guan840912 Wondering why you closed this one? Was actually planning on taking a look this week.

guan840912:
I think that no one has reviewed it, maybe I was the only one who encountered this problem, so I closed it

packages/cdk8s-plus-17/src/ingress-v1beta1.ts
```diff
@@ -219,6 +219,17 @@ export class IngressV1Beta1Backend {
...
   }
+  /**
+   * Redirect Traffic from HTTP to HTTPS At AWS LoadBalancer Controller,
+   * Using the alb.ingress.kubernetes.io/actions.${action-name} annotation to setup an ingress to redirect http traffic into https
+   * and return ssl Readirect rule to Ingress.
+   */
+  public static sslRedirect() {
+    return new IngressV1Beta1Backend({
+      serviceName: 'ssl-redirect',
+      servicePort: 'use-annotation',
```
iliapolo, line 230: I'm not sure I understand - are these well known values for this type of backend? What makes this specific for SSL redirect?

iliapolo: Ok I see they are used in the example you referenced to: 

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  namespace: default
  name: ingress
  annotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/certificate-arn: arn:aws:acm:us-west-2:xxxx:certificate/xxxxxx
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS":443}]'
    alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'
spec:
  rules:
    - http:
        paths:
         - path: /*
           backend:
             serviceName: ssl-redirect
             servicePort: use-annotation
         - path: /users/*
           backend:
             serviceName: user-service
             servicePort: 80
         - path: /*
           backend:
             serviceName: default-service
             servicePort: 80
```

My questions are then:

- Is `use-annotation` a reserved name for a service port in k8s? not familiar with it.
- is `ssl-redirect` a reserved action? or can it be some other string that just needs to match the annotation of the ingress?

guan840912: it is only for` AWS LoadBalancer Controller` `Redirect Traffic from HTTP to HTTPS`.


It'II use the `alb.ingress.kubernetes.io/actions.${action-name}` annotation to setup an ingress to redirect http traffic into https
like
`alb.ingress.kubernetes.io/actions.ssl-redirect: '{"Type": "redirect", "RedirectConfig": { "Protocol": "HTTPS", "Port": "443", "StatusCode": "HTTP_301"}}'`

more details please see https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/tasks/ssl_redirect/ ,thanks. 😊

iliapolo: Ok I understand. The problem with providing an `sslRedirect` method like you suggested is that:

- It will not work if the ingress doesn't have the correct annotations.
- It will not work if the ingress uses a different action name than `ssl-redirect`.

For it too be more robust and clear, It would need to look something like this:

```ts
IngressV1Beta1Backend.fromAwsAlbAction('ssl-redirect');
```

Which makes it somewhat clear that this relates to an `ssl-redirect` action that must be configured as an ingress annotation. But at this point this only saves the user from having to configure `use-annotation` as the service port.
Given that, I think it is too specific and error prone to provide the desired benefit. 

What I would do is actually create a new kind of `Ingress`:

```ts
const ingress = new kplus.AwsALBIngress(this, 'Ingress', {
  sslRedirect: true,
});
```

Which would automatically configure both the annotations and the backend. We can also add some nice API for this to add additional actions:

```ts
ingress.addAction('/some-path', someAction);
```

This is a perfect use-case for the intent based API's we are targeting with `cdk8s-plus`. 

If you're up for a giving a go this that would be great and we can discuss how exactly to implement it.

guan840912: Thank you for your explanation, I am willing to implement a kplus.AwsALBIngress() class. 😊

iliapolo: @guan840912 would you be interested in publishing this ingress as a separate module? I think that would be the best approach for now, we can have a link in our README pointing to your package so people can easily discover it :)

guan840912: Okay, I will close this PR, create another project for this `AwsALBIngress()`;)

iliapolo: Awesome - have a look at [`cdk8s-image`](https://github.com/eladb/cdk8s-image) as an example and [`projen`](https://github.com/projen/projen) that will help with the project configuration and release workflows 👍

PR #514, opened 2021-01-21
chore(release): 1.0.0-beta.6

iliapolo:


eladb (approved):

iliapolo:
@eladb The build is failing because the integration test snapshots are now missing the imports of:

- apiextensions.crossplane.io.ts
- meta.pkg.crossplane.io.ts
- pkg.crossplane.io.ts

Apparently this no longer takes affect:

https://github.com/awslabs/cdk8s/blob/a49a6b187875cc0f52fc6ad3261f211cf2416d80/packages/cdk8s-cli/src/import/crd.ts#L127-L129

The CRD's themselves are hosted here: https://doc.crds.dev/raw/github.com/crossplane/crossplane@v0.14.0.

And indeed they do not have a top level `kind` property. It seems like the objects now only include the `spec`. Do you have any idea? Maybe @prasek can help out?

iliapolo:
I think this might be related to: https://github.com/crdsdev/doc/commit/fee7e2b090e8f6e507a8af5f3c1aa08e64bb2267#diff-d3a059d47a8abdf6c9b6971f64d4ec9089ef3ac63bb2de55089fba18832f67b7

hasheddan:
@iliapolo apologies for the break here. This has to do with the fact that CRDs are now being stored with internal representation, so those metadata fields are stripped off, but all of the same information is technically available. That endpoint could be updated to return with `v1` schema if needed. As a side note, all repos are now indexed by https://doc.crds.dev so this should be much more generally useful now :)

iliapolo:
@hasheddan So all of the objects currently returned by the endpoint will always have the same GVK?

hasheddan:
@iliapolo yes that is correct, it will _only_ return `kind: CustomResourceDefinition`. However, it is the full internal spec right now. I have been talking with @prasek and am thinking of making the `raw/` endpoint just return `v1` flavor of CRDs though so that will be easier for y'all to parse. For instance, with my changes, you would get something like this again:

```
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  creationTimestamp: null
  name: configurationrevisions.pkg.crossplane.io
spec:
  group: pkg.crossplane.io
  names:
    categories:
    - crossplane
    kind: ConfigurationRevision
    listKind: ConfigurationRevisionList
    plural: configurationrevisions
    singular: configurationrevision
  scope: Cluster
  versions:
...
```

I assume this will be most convenient for you? If not, happy to support other variations :)

iliapolo:
@hasheddan yes that would be great. I also think it would be more aligned with standard k8s resource definitions, making it easier for other tools to consume as well. 

Thanks!

hasheddan:
@iliapolo great :) Should be fixed https://doc.crds.dev/raw/github.com/crossplane/crossplane

hasheddan:
Feel free to tag me on anything if you are seeing issues in the future :+1:

iliapolo:
@hasheddan sweet, thanks! :)

PR #520, opened 2021-01-26
chore:  update cdk8s version

curry9999:
This PR does 2 things:
- The cdk8s version up. 0.0.0 to 1.0.0-beta.6.
- Because an error occurs in cdk8s version 0.0.0
- Fixed a README typo


examples/typescript/cdk8s-plus-elasticsearch-query/package.json
```diff
@@ -17,14 +17,14 @@
     "upgrade:next": "npm i cdk8s@next cdk8s-cli@next"
   },
   "dependencies": {
-    "cdk8s": "0.0.0",
-    "cdk8s-plus-17": "0.0.0",
+    "cdk8s": "1.0.0-beta.6",
```
campionfellin, line 20: These are intentionally `0.0.0` so that we can align the versions on release (basically grep for that value and replace). (script [here](https://github.com/awslabs/cdk8s/blob/ffce8c696ce502ec4a14f65b715474faa7f74c7b/tools/align-version.sh)).

Is this causing a bug for you?

PR #528, opened 2021-02-05
feat: Allow configuration of how YAML output is divided into files

### LINKED ISSUE #115: [Suggestion] Synth files per resource instead files per chart
It would be cool to have the option to synth the manifests per resource instead of per chart. 
It could be a dump of all manifests, or even create folders per chart and place the manifests in there.

I would propose that we add `outputYamlFormat` option in the `cdk8s.yaml` config file that accepts these values:
- `filePerChart`: Current behavior - defaults to this if not specified
- `folderPerChart`: Creates a folder with the **Chart** name and writes the **Resources** such as `deployment.yaml`, `services.yaml`, and so on. 
- `filePerResource`: dump of a file for each resource on the `dist` folder

For the `folderPerChart` option we could assume the chart name as the folder name and for the filePerResource, we can assume some convention, like `chartName-resourcekind`, e.g. `myapp-deployment`, `my-app-service` and so on.
If there are multiple resources of the same kind, they can stay within the same `yaml` file as now to avoid file naming problems.

Thoughts?

heckler1:
This adds an optional configuration value to AppProps, to allow the output from the App to be divided into files in a couple different ways:

- One YAML file for the entire App
- One YAML file per chart, which is the current behavior, and the new default
- One YAML file per K8S resource, with the file name formatted as `${ApiVersion}.${Kind}.${Name}.k8s.yaml`

This should be backwards compatible, as the per chart behavior has not changed.

I still need to write tests for the non-default modes, but as it stands right now, the current tests pass locally. I'm open to suggestions on a better way of implementing this, it may not be as intuitive as I think it is.

Fixes #115 

---

### DIFF
--- a/packages/cdk8s/API.md
+++ b/packages/cdk8s/API.md
@@ -49,6 +49,7 @@ Name|Description
 Name|Description
 ----|-----------
 [SizeRoundingBehavior](#cdk8s-sizeroundingbehavior)|Rounding behaviour when converting between units of `Size`.
+[YamlOutputType](#cdk8s-yamloutputtype)|The way to divide YAML output into files.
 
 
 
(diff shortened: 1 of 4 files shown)


iliapolo:
@heckler1 Sorry for the long delay. I'm happy to take a look at this but we've actually changed our repo structure recently. 

Any chance you can resubmit this PR against the https://github.com/cdk8s-team/cdk8s-core repo?

heckler1:
@iliapolo I've opened https://github.com/cdk8s-team/cdk8s-core/pull/24. Just let me know if there's anything else I can do 👍

PR #454, opened 2020-12-15
feat(cli): new version notifications

eladb:
The CLI will print a notification if there is a new version of the CDK8s CLI available in npm.

The result is cached under ~/.cdk8s-cli.version with a TTL of 30min.

Resolves #452

---

Screen Shot 2020-12-15 at 23 06 38

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*

Issue #363, opened 2020-11-05
[cdk8s] Getting ENOBUFS when including a larger than normal manifest file


van-vothanh:
Hello,

I'm running into an issue with my `cdk` and `cdk8s` (including `cdk8s+`) app when including a larger than normal manifest file.

Basically, what I'm trying to do is install the new `aws-load-balancer-controller` on my `EKS` cluster using `cdk8s`. As per this step (https://kubernetes-sigs.github.io/aws-load-balancer-controller/guide/controller/installation/#install-cert-manager) of the instructions, I need to install the `cert-manager` add-on, however as soon as I include its manifest file (https://github.com/jetstack/cert-manager/releases/download/v1.0.2/cert-manager.yaml), which is over 1.7MB, I receive an `ENOBUFS` error upon `cdk synth` or `cdk deploy`.

```typescript
new cdk8s.Include(this, `certificate-manager`, {
    url: 'https://github.com/jetstack/cert-manager/releases/download/v1.0.2/cert-manager.yaml'
});
```

Has anyone come across this issue and if so how did you guys get around it?

I did do a bit of digging and realised that the manifest file's size is a tad over the default nodejs's `child_process`'s `maxBuffer` of 1MB, and `cdk` currently doesn't give us the options to override this value (as per https://github.com/aws/aws-cdk/blob/5dcdecb2c5d6ce19517af66090cfacabed88025b/packages/aws-cdk/lib/api/cxapp/exec.ts#L106). Having said that, I'm not 100% sure if this is the culprit.

Thanks

danikenan:
here is a solution. If allowed, will improve and make a pull request:

```
import { ApiObject, IncludeOptions } from 'cdk8s';
import { Construct, Node } from 'constructs';
import { resolve } from 'path';
import { existsSync, readFileSync, writeFileSync } from 'fs';
import request from 'sync-request';
import { safeLoadAll } from 'js-yaml';
import { tmpdir } from 'os';

export function load(url: string, cacheDir: string | false = tmpdir()) {
  const fileName = url.replace(/^.*\/([^/]+)$/, '$1');
  const cacheFileName = cacheDir ? resolve(cacheDir, fileName) : undefined;
  const yml =
    cacheFileName && existsSync(cacheFileName)
      ? readFileSync(cacheFileName, 'utf-8')
      : request('GET', url).getBody().toString('utf-8');
  if (cacheFileName && !existsSync(cacheFileName)) {
[...]

PR #529, opened 2021-02-06
fix(lib): ENOBUFS for large helm charts

dave-b-code:

With charts such as cert-manager with crds included, the helm import functionality can run into an ENOBUFS error due to insufficient buffer size. This is very similar to issue #363 (except with helm instead of yaml download) and uses a very similar fix to PR #454. Thanks.


eladb (approved):
Looks good. Thanks

PR #543, opened 2021-03-03
feat: add flag to control upgrade check

vbatoufflet:
Hi,

This PR intends to define a new flag on the cdk8s-cli (i.e. `--check-upgrade`) to control available upgrade check trigger while running the command.

Regards

### DIFF
--- a/packages/cdk8s-cli/src/cli/index.ts
+++ b/packages/cdk8s-cli/src/cli/index.ts
@@ -3,16 +3,22 @@ import * as yargs from 'yargs';
 import { upgradeAvailable } from '../upgrades';
 
 async function main() {
-  const versions = upgradeAvailable();
-  if (versions) {
-    console.error('------------------------------------------------------------------------------------------------');
-    console.error(colors.yellow(`A new version ${versions.latest} of cdk8s-cli is available (current ${versions.current}).`));
-    console.error(colors.yellow('Run "npm install -g cdk8s-cli" to install the latest version on your system.'));
-    console.error(colors.yellow('For additional installation methods, see https://cdk8s.io/docs/latest/getting-started'));
-    console.error('------------------------------------------------------------------------------------------------');
-  }
-
   const ya = yargs
+    .option("check-upgrade", {type: "boolean", desc: "Check for cdk8s-cli upgrade", default: true})
+    .check(argv => {
+      if(argv.checkUpgrade) {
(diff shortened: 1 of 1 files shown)


rossi-fi:
Thanks for this! My only question is how this translates into an environmental variable. I eventually got it working while not changing the command line option, like this:

`.option("check-upgrade", {type: "boolean", default: true, desc: "Check for cdk8s-cli upgrade (default true)"})`

Now I can set `CDK8S_CHECK_UPGRADE=false` as the default value in eg. a container.

packages/cdk8s-cli/src/cli/index.ts
```diff
@@ -3,16 +3,22 @@ import * as yargs from 'yargs';
...
-  if (versions) {
-    console.error('------------------------------------------------------------------------------------------------');
-    console.error(colors.yellow(`A new version ${versions.latest} of cdk8s-cli is available (current ${versions.current}).`));
-    console.error(colors.yellow('Run "npm install -g cdk8s-cli" to install the latest version on your system.'));
-    console.error(colors.yellow('For additional installation methods, see https://cdk8s.io/docs/latest/getting-started'));
-    console.error('------------------------------------------------------------------------------------------------');
-  }
-
   const ya = yargs
+    .option("no-check-upgrade", {type: "boolean", desc: "Do not check for cdk8s-cli upgrade"})
```
rossi-fi, line 7: Please consider using `check-upgrade` instead of `no-check-upgrade` and set default to true. This way you can still do `--no-check-upgrade` and also use CDK8S_CHECK_UPGRADE=false.

vbatoufflet: Good idea. I'll do that 👍

vbatoufflet: Just updated the PR to implement those changes.

rossi-fi (approved):
Thank you for this!

eladb:
@vbatoufflet  apologies for the inconvenience! We have moved this module to https://github.com/cdk8s-team/cdk8s-cli. Could you please re-submit this pull request against the new repository?

vbatoufflet:
> Could you please re-submit this pull request against the new repository?

@eladb sure no worries, I'll re-submit on the new repository 👍

eladb:
Thanks, and apologies again for the burden.

PR #549, opened 2021-03-17
fix(docs): correct language tab indentation

outsideris:

I just found [a broken doc for import](https://cdk8s.io/docs/v1.0.0-beta.10/cli/import/#module-name) when I read the page. 

It is because of the wrong indentation for the code blocks.

I checked it rendered correctly in my local.


eladb (approved):
🙏

PR #590, opened 2021-05-27
chore: delete modules and use latest from npm

eladb:
Deletes the code of everything under `packages/**` and takes a dependency on the published modules instead. Modules are new mastered in [cdk8s-core], [cdk8s-cli] and [cdk8s-plus-17].

[cdk8s-core]: https://github.com/cdk8s-team/cdk8s-core
[cdk8s-cli]: https://github.com/cdk8s-team/cdk8s-cli
[cdk8s-plus-17]: https://github.com/cdk8s-team/cdk8s-plus-17

Doc website, examples and integration tests are now executed against the released versions (installed via the root `package.json` file).

---


.github/workflows/build.yml
```diff
@@ -1,5 +1,5 @@
 name: Build
-on: [pull_request]
+on: [push]
```
iliapolo, line 2: Why the change?

examples/java/crd/package.json
```diff
@@ -7,12 +7,5 @@
     "build": "cdk8s import --language java",
     "synth": "mvn install && mvn compile && cdk8s synth",
     "test": "echo 'no test for java yet.'"
-  },
-  "dependencies": {
```
iliapolo, line 11: Can't we leave these with the current latest version and let the dependency upgrade workflow take care of upgrading it? (i'm not sure why it didn't already...)

iliapolo (commented):
@eladb Maybe add a README note for a while saying the packages were moved? I would also post something on our slack channel notifying users of this change.

.github/workflows/build.yml
eladb: I want this workflow to run against any push (both PRs and pushes to master)

examples/java/crd/package.json
eladb: It's much easier to manage this at the root level I believe. I'll also revisit the upgrade workflow shortly (via projen) after this first iteration is complete

eladb:
@iliapolo Ready for another round

iliapolo (approved):

PR #669, opened 2021-09-01
feat: integration tests for all platforms

### LINKED ISSUE #422: Windows support
We should officially support windows before announcing GA. This means adding some windows CI in GitHub actions as well.

* [ ] :wave: I may be able to implement this feature request

---

Chriscbr:
Closes #422

----


Chriscbr:
We need the changes from https://github.com/cdk8s-team/cdk8s-cli/pull/63 to be merged before any windows tests will pass.

Chriscbr:
Integration tests are passing now. I've set them up so that the `build` task runs in parallel with three `test` tasks that each run the integration suite on a different platform.

Since we are running additional tasks in addition to `build` that each could fail independently, we don't want any snapshots to be updated during the build, since that would cause projen to generate a new self-mutation commit that would automatically be approved. Thus, I've had to change the `build` and `test` tasks to run without updating snapshots automatically. (But we can still run `test:update` and `integ:update` locally).

I think it's possible we could bring back snapshot updating in the `build` task by adding an option to projen that disables self-mutation commits, but this is the best workaround I could come up with for now.

.github/workflows/build.yml
```diff
@@ -53,3 +53,30 @@ jobs:
...
+      - name: Install pipenv
+        run: pip install pipenv
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.16"
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Run integration tests
+        run: yarn run integ
```
iliapolo, line 75: Currently the integ tests are running as part of the build task:

https://github.com/cdk8s-team/cdk8s/blob/7ae13bffc149f4770ff6848665578dcbd2103c88/.projen/tasks.json#L52-L58

How come this doesn't need a `lerna run build` before?

.github/workflows/build.yml
```diff
@@ -53,3 +53,30 @@ jobs:
           }}/actions/runs/${{ github.run_id }}/cancel
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  test:
```
iliapolo, line 56: In this setup, we actually run ubuntu integ tests twice - once in the `build` task and once here. There are 3 options here:

1. Reduce the matrix to only `windows-latest` and `macos-latest`. Dont really like this since it breaks symmetry. 
2. Remove the `integ` task from `build`. Dont really like this since it means running `npx projen build` locally will not run integ tests.
3. Patch the build job with a matrix. I think this is an elegant solution since it keeps everything in standard form and also runs the unit tests on all platforms (i guess this could be a problem if our unit tests are platform dependent, but i don't think they are, or should be). This is a feature I think should eventually go into projen, where the self-mutation step would run only after all the matrix jobs have passed.

Thoughts?

.github/workflows/build.yml
Chriscbr: We don't need to run a build for the integ tests because they don't rely on any source code within this repo (they're all bash scripts).

.github/workflows/build.yml
Chriscbr: Discussed offline - I've updated the integ tests to run in a separate workflow, which should resolve the self-mutation issues. I think it's okay that the ubuntu integ test still runs twice, since the one in the `build` workflow has snapshot updating enabled while the ones in `integ` don't. But I'm totally open to changing it.

iliapolo:
@Chriscbr The new workflow needs to be added as a required status check to mergify:

https://github.com/cdk8s-team/cdk8s/blob/rybickic/integ/.mergify.yml#L15

I really hope mergify will accept the os suffixes to job names...

As well as in the github branch protection (manually for now).

iliapolo (changes requested):
@Chriscbr The new workflow needs to be added as a required status check to mergify:

https://github.com/cdk8s-team/cdk8s/blob/rybickic/integ/.mergify.yml#L15

I really hope mergify will accept the os suffixes to job names...

As well as in the github branch protection (manually for now).

iliapolo (approved):

PR #707, opened 2021-10-06
chore(docs): relicensing docs to CC-BY-4.0

hyandell:
I've reviewed contributions to the docs and am relicensing them in accordance with CNCF IP policy over to CC-BY-4.0.

I've also updated the website footer to mention this, and to fix the Trademark link. These both match the Kubernetes website, which I am treating as the gold standard. 

And I've removed the pull request template with Apache-2.0 language; DCO does not typically rely on an additional statement in pull requests on licensing, this matches Kubernetes, and it saves figuring out how to fit CC-BY-4.0 and Apache-2.0 together there.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,1 +1,0 @@
-*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license*


iliapolo (changes requested):
@hyandell The PR template is actually managed via projen. And needs to be removed here:

https://github.com/cdk8s-team/cdk8s/blob/master/.projenrc.js#L9

Make sure to also add `pullRequestTemplate: false` and run `yarn build` to regenerate the files.

I'm happy to take this over if needed. let me know.

PR #715, opened 2021-10-11
chore: delete legacy workflows

eladb:
Remove workflows that are not managed by projen.


iliapolo:
@eladb this needs DCO sign-off

PR #750, opened 2021-11-11
fix(docs): undefined id error on index doc Go section

camba1:
As originally written, the Go example in the index page throws an error because id is undefined. This PR fixes the document by defining id and setting it equal to "ubuntu" (since we are creating an ubuntu deployment)

### DIFF
--- a/docs/plus/index.md
+++ b/docs/plus/index.md
@@ -270,6 +270,7 @@ app.synth();
 
     ```go
     app := cdk8s.NewApp(nil)
+    id := "ubuntu"
     chart := cdk8s.NewChart(app, jsii.String(id), nil)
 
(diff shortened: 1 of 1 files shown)


docs/plus/index.md
```diff
@@ -270,6 +270,7 @@ app.synth();
 
     ```go
     app := cdk8s.NewApp(nil)
+    id := "ubuntu"
     chart := cdk8s.NewChart(app, jsii.String(id), nil)
```
eladb, line 273: Wouldn't it make more sense to just use a string literal instead of assigning to a variable (get rid of `id` altogether)?

camba1: Sounds good. I have updated the doc to use the literal directly

eladb (approved):

PR #785, opened 2021-12-10
feat: New yaml output type: Folder_per_chart_file_per_resource 

camba1:
This PR refers to documenting the changes from[ PR 143 ](https://github.com/cdk8s-team/cdk8s-core/pull/143 )in the cdk8s-core repo , which adds a new yaml output type:` folder_per_chart_file_per_resource`.

The new type allows users to elect to export their yaml resources as follows:

- Each chart creates a folder
- Each resource is created as a file and placed in the correct chart folder
This provides better organization and flexibility as the number of resources created increases

### DIFF
--- a/docs/reference/cdk8s/java.md
+++ b/docs/reference/cdk8s/java.md
@@ -2156,3 +2156,8 @@ Each resource is output to its own file.
 
 ---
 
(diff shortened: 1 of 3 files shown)


Chriscbr:
The docs generated here are automatically generated by [jsii-docgen](https://github.com/cdklabs/jsii-docgen) by a GitHub workflow we have, so there's no need to update this by hand. 🙂

Closing.

PR #792, opened 2021-12-17
chore: add k8s 1.23.0 schema

Chriscbr:


Chriscbr:
I expect the build fail due to issues publishing cdk8s-core the past several days, but it should still be safe to merge this.

PR #804, opened 2021-12-29
chore(deps): upgrade dependencies

Chriscbr:
Signed-off-by: Christopher Rybicki <rybickic@amazon.com>

PR #811, opened 2022-01-10
chore: refactor integ tests to use new projen apis

Chriscbr:
Supersedes #804

This updates `.projenrc.js` to avoid having to manually override all of `mergify.yml`.


Chriscbr:
Merging this in since the changes were already approved in #804.

(There is a mergify build check failing, but it's because Mergify deprecated "Strict mode" today - this will be fixed upstream in projen)

PR #828, opened 2022-01-28
chore: remove outdated API reference files

Chriscbr:
These files make it look in the repo like our API reference is outdated, but the files are really just being generated during the website release process (and not being committed to the repo). 


madeline-k (approved):
😄

PR #889, opened 2022-03-27
chore: docs for pv & pvc

iliapolo:
Docs for https://github.com/cdk8s-team/cdk8s-plus/pull/483


docs/plus/pv.md
```diff
@@ -0,0 +1,76 @@
...
+  // assign the volume to small-ebs storage class
+  storageClassName: 'small-ebs',
+
+  // what is the volume storage
+  storage: cdk8s.Size.gibibytes(50),
+});
+```
+
+Note that this **does not** actually create a new volume, it merely manifests an existing
+volume in AWS as a Kubernetes resource.
```
Chriscbr, line 38: Good callout

docs/plus/pv.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+## Bind
+
+Binding is a part of the reservation process, but it only creates a one directional link.
+You can use it to bind a PV to an existing PVC. Note however that if the PVC is not bound to the PV,
+there's no guarantee this volume will indeed be given that specific claim.
+
+```ts
+const claim = kplus.PersistentVolumeClaim.fromClaimName('claim');
```
Chriscbr, line 71: Would it make sense to also have an API to create a PVC from a selector, e.g.:

```ts
const claim = kplus.PersistentVolumeClaim.fromSelector(...)
```

(I'm open to leaving this as a future enhancement)

Chriscbr (approved):
Nice docs 👍  Approving with do-not-merge so it's not merged in before the implementation PR.

Also cc'ing @misterjoshua in case he has any feedback on the API/docs, since I think he was writing a PR to support PVCs not too long ago? :)

docs/plus/pv.md
```diff
@@ -0,0 +1,76 @@
...
+
+```ts
+container.mount('/data', kplus.Volume.fromPersistentVolumeClaim(claim));
+```
+
+You can also directly mount a persistent volume, which will implicitly reserve it:
+
+```ts
+container.mount('/data', kplus.Volume.fromPersistentVolume(vol));
+```
```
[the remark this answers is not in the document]
iliapolo: I agree its a bit confusing. Added a note that hopefully clears it up a bit. (https://stackoverflow.com/a/5450075 points to some UI question, guess you meant a different link?).

> I wonder if we should just be able to pass vol directly (since "Volume.fromXxxVolume" sounds redundant at first glance)?

Good idea, it required some rewiring but I think I came up with something reasonable. Let me know what you think.
Unfortunately the original PR was already merged by the time I got to it, but here is a new one: https://github.com/cdk8s-team/cdk8s-plus/pull/491

docs/plus/pvc.md
```diff
@@ -0,0 +1,77 @@
...
+import * as kplus from 'cdk8s-plus-22';
+import * as cdk8s from 'cdk8s';
+
+const pod = new kplus.Pod(chart, 'Pod');
+const container = pod.addContainer({ image: 'node' });
+
+// create the storage request
+const claim = new kplus.PersistentVolumeClaim(chart, 'Claim', {
+  storage: cdk8s.Size.gibibytes(50),
+});
```
misterjoshua, line 20: How do we envision supporting CSI-based [PVC cloning](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#create-persistent-volume-claim-from-an-existing-pvc) and [PVC from Volume Snapshot](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-snapshot-and-restore-volume-from-snapshot-support)?

How about something like this?

```ts
const claimFromClone = new kplus.PersistentVolumeClaim(chart, 'Claim', {
  dataSource: PersistentVolumeClaimDataSource.pvcDataSource(otherClaim),
});

const claimFromSnapshot = new kplus.PersistentVolumeClaim(chart, 'Claim', {
  dataSource: PersistentVolumeClaimDataSource.snapshotDataSource({
    name: 'new-snapshot-test',
    kind: 'VolumeSnapshot',
    apiGroup: 'snapshot.storage.k8s.io',
  }),
});
```

misterjoshua (commented):
@Chriscbr Thanks for pinging me! I would have missed this PR otherwise. :)

This PR looks pretty good to me. I've asked a few questions on supporting `dataSource` for CSI storage drivers.

docs/plus/pv.md
Chriscbr: Sounds good 👍

docs/plus/pv.md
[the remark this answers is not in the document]
Chriscbr: Whoops, the URL should have been https://stackoverflow.com/a/54500752 ("What is the difference between a volume and persistent volume?")

The updated API still looks good, though I'm also wondering if it will be confusing that you can do:

```ts
container.mount('/data', pv);
```

which implicitly creates a PVC for you and mounts it -- but if you have a PVC already you need to do:

```ts
container.mount('/data', Volume.fromPersistentVolumeClaim(pvc))
```

What do you think of either:
- Make PVC implement `IStorage` (I don't think this mental model makes sense)
- Replace`IStorage` with `IMountable` and make `mount()` accept any `IMountable` (which could be either a Volume or PVC -- or possibly PV).
- Keep the existing API which makes the volume sources always explicit (drop this PR)

Chriscbr (approved):

docs/plus/pv.md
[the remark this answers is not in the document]
iliapolo: As discussed offline, we decided its ok to have `mount` accept a PV, but not a PVC. The mental model of users is mounting a volume, not a claim, which is somewhat of an implementation details for this API. 

We'll keep this API for now and reconsider at a later point if needed.

docs/plus/pvc.md
iliapolo: @misterjoshua yeah I think you're proposed API is the right direction. We will create a union like class called `PersistentVolumeClaimDataSource` that will offer all these data sources 👍 

We can add it as an enhancement later on, do you wanna have a go at adding it? :)

PR #911, opened 2022-05-06
chore: docs for scheduling

iliapolo:
Docs for https://github.com/cdk8s-team/cdk8s-plus/pull/684 (should be merged after)

Signed-off-by: iliapolo <epolon@amazon.com>


docs/plus/deployment.md
```diff
@@ -75,3 +75,63 @@ spec:
...
+
+This example ensures that each replica of the `Redis` deployment
+will be scheduled on a different node.
+
+Take, for [example](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#more-practical-use-cases), a three-node cluster running a web application with an in-memory cache like redis. You'd like to co-locate the web servers with the cache as much as possible, while still maintaining node failure resistance. (i.e not all pods are on the same node).
+
+Here is how you can accomplish that:
+
+```ts
+import * as k from 'cdk8s-plus-22';
```
Chriscbr, line 114:
```suggestion
import * as k from 'cdk8s';
```

docs/plus/pod.md
```diff
@@ -57,4 +57,208 @@ const chart = new k.Chart(app, 'Chart');
...
+the `Redis` pod. To request a **preference**, specify the [`weight`](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity-weight) property:
+
+```ts
+web.scheduling.colocate(redis, { weight: 50 });
+```
+
+To use a different topology, specify the `topologyKey` property:
+
+```ts
+web.scheduling.colocate(redis, { weight: 50, topologyKey: kplus.TopologyKey.ZONE });
```
Chriscbr, line 192: WDYT of:
```suggestion
web.scheduling.colocate(redis, { weight: 50, topology: kplus.Topology.ZONE });
```

Chriscbr (approved):
Looks great Eli! 😄

docs/plus/deployment.md
```diff
@@ -75,3 +75,63 @@ spec:
...
+### Spreading
+
+A spread is a [separation](./pod.md#pod-separation) of pods from themselves.
+It can be used to ensure replicas of the same workload are scheduled on different topologies.
+
+> The same API is also available on all workload resources (i.e `Deployment`, `StatefulSet`, `Job`, `DaemonSet`).
+
+```ts
+import * as k from 'cdk8s-plus-22';
+import * as kplus from 'cdk8s-plus-22';
```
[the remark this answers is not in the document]
iliapolo: Yeap, thanks :)

docs/plus/pod.md
```diff
@@ -57,4 +57,208 @@ const chart = new k.Chart(app, 'Chart');
...
+import * as kplus from 'cdk8s-plus-22';
+
+const app = new k.App();
+const chart = new k.Chart(app, 'Chart');
+
+const redis = new kplus.Pod(chart, 'Redis', {
+  containers: [{ image: 'redis' }]
+});
+const highMemoryNodes = kplus.Node.select(kplus.NodeLabelQuery.is('memory', 'high'));
+redis.scheduling.attract(highMemoryNodes);
```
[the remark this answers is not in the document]
iliapolo: I'm having some doubts about the name though. Even though it is used in the k8s docs:

> [Node affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) is a property of [Pods](https://kubernetes.io/docs/concepts/workloads/pods/) that attracts them to a set of [nodes](https://kubernetes.io/docs/concepts/architecture/nodes/) (either as a preference or a hard requirement). Taints are the opposite -- they allow a node to repel a set of pods.

The word `attract` doesn't convey a **requirement** by default, but actually feels more of a preference. This is reversed to how the API is built, where the default is the requirement, and the preference needs to be explicitly specified. Can you think of another word? 

BTW the other method names don't suffer from this (i.e `assign`, `colocate`, and `separate` all convey a requirement, I feel)

docs/plus/pod.md
```diff
@@ -57,4 +57,208 @@ const chart = new k.Chart(app, 'Chart');
...
+import * as kplus from 'cdk8s-plus-22';
+
+const app = new k.App();
+const chart = new k.Chart(app, 'Chart');
+
+const redis = new kplus.Pod(chart, 'Redis', {
+  containers: [{ image: 'redis' }]
+});
+
+redis.scheduling.tolerate(kplus.Toleration.noSchedule('key1', 'value1'));
```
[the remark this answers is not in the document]
iliapolo: Yeah you're right, I was going back and forth with this API, but I agree that using the effect as the function name conveys the wrong intent. I was trying to find a way to make `tolerate` accept a `Node`, but wasn't able to.

Your last suggestion is exactly what I wanted! :) It also provides good symmetry with the other APIs. 
So I ended up going with:

```ts
const devNodes = kplus.Node.tainted(kplus.NodeTaintQuery.exists('group', { effect: TaintEffect.NO_SCHEDULE }));
```

Let me know what you think, Thanks!

docs/plus/pod.md
iliapolo: Yeap, I like it 👍

Chriscbr (approved):

docs/plus/pod.md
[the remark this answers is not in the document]
Chriscbr: `link`? `bond`?

Chriscbr (approved):

docs/plus/pod.md
```diff
@@ -57,4 +93,267 @@ const chart = new k.Chart(app, 'Chart');
...
+
+Pod selection can also be scoped to specific namespaces.
+This is done using the `.namespaced` method, which can accept any [namespace selector](./namespace.md#select-a-namespaces).
+
+For example, select all pods that have the `app=store` label in the `backoffice` namespace:
+
+```ts
+const pods = kplus.Pod.labeled(kplus.LabelQuery.is('app', 'store'))
+  .namespaced(kplus.Namespace.named('backoffice'));
+```
```
Chriscbr, line 121: What do you think of putting these static methods in a separate `Pods` class to emphasize that a selection could contain any number of pods (zero or more)? I'm wondering if this could possibly match the intent of this line of code a little better: "selecting all `Pods` `labeled` with `app=store` and `namespaced` under `backoffice`".

This would also make some of the other class names a little bit clearer, e.g. `LabeledPod` -> `LabeledPods` makes it more clear that `LabeledPod` isn't just a specific type of pod. :-)

Chriscbr, line 121: I'm curious if switching the `Pod.select()`-based API with a chaining-based API was mostly an ergonomics choice, or if there's any other benefits we get? I remember some folks saying the chaining APIs are awkward in some JSII languages, though I don't want that to necessarily dictate the design here. 

I have a slight bias towards the non-chaining version, since I reckon it might be easier to extend with more options if necessary in the future - but I'm interested in hearing your perspective 🙂. I also noticed the chaining implementation imposes an order on the fields (you have to specify labels and then namespaces) - I wonder if that could become awkward? Maybe we could offer both styles? 
```ts
const pods = kplus.Pods.select({
  selectors: [kplus.LabelQuery.is('app', 'store')],
  namespace: kplus.Namespace.named('backoffice'),
});
// or
const pods = kplus.PodSelection.of({ ... })
```

docs/plus/pod.md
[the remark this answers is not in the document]
iliapolo: Nah, i'll leave `attract` for now.

docs/plus/pod.md
iliapolo: Yeah I agree, will change.

iliapolo: I gave some more though after our conversation and decided you're right. 

The type-safety suggested by my API complicates the ergonomics too much, especially when network policies are introduced, which has a different selection heuristic. I went with `Pods.select`, and we will add runtime checks where necessary

PR #920, opened 2022-05-13
fix: docs refer to non existent `EnvValue.fromSecret` function

oura-kataja:
The API does not have the function `kplus.EnvValue.fromSecret` which was used in example. Changed it to use `kplus.EnvValue.fromSecretValue` instead.


oura-kataja:
Seeing that  cdk8s-team/cdk8s-plus#688 removes `addEnv(...)` and replaces that with `env.addVariable(...)`  and `Env.fromSecret(...)`, it's unclear now that translates to env variables. I don't understand the reasons behind the changes. Closing the PR.

iliapolo (approved):
Thanks!

PR #924, opened 2022-05-22
feat(kubernetes-schemas): schema for `1.24.0`

iliapolo:
Signed-off-by: iliapolo <epolon@amazon.com>


avaly:
Why are the schema definitions starting with 1.24 missing enum values?

Here's the 1.23 definitions file vs the 1.24 definitions file:

image

PR #952, opened 2022-07-13
chore(website): `CronJob` L2 docs

vinayak-kukreja:
CDK8s+ CronJob Readme

Signed-off-by: Vinayak Kukreja <vinakuk@amazon.com>


docs/plus/cronjob.md
```diff
@@ -0,0 +1,93 @@
+# CronJob
+
+CronJob resource is responsible for creating recurring [Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/). The job recurrence is determined by a [Schedule]() provided. You can also specify a custom schedule by passing a [cron pattern](https://en.wikipedia.org/wiki/Cron) to Schedule's cron method.
+
```
iliapolo, line 4: Add a link to the future API Reference page, like here:

https://github.com/cdk8s-team/cdk8s/blob/ade2e3de0d8b7193e09804f5fbf37c5f6bada7d5/docs/plus/config-map.md?plain=1#L6-L7

docs/plus/cronjob.md
```diff
@@ -0,0 +1,93 @@
+# CronJob
+
+CronJob resource is responsible for creating recurring [Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/). The job recurrence is determined by a [Schedule]() provided. You can also specify a custom schedule by passing a [cron pattern](https://en.wikipedia.org/wiki/Cron) to Schedule's cron method.
```
iliapolo, line 3:
```suggestion
`CronJob` resource is responsible for creating recurring [Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/). The job recurrence is determined by a provided [Schedule]().
```

docs/plus/cronjob.md
```diff
@@ -0,0 +1,93 @@
...
+
+CronJob is similar to a job but it is suitable when there is a need to run a job indefinitely following a schedule. These repitive jobs can be utilized for recurring tasks like backing up a database, pinging a server for health checks, creating snapshots of system and much more.
```
iliapolo, line 5:
```suggestion
CronJob is similar to a [job](https://cdk8s.io/docs/latest/plus/job/) but is suitable when there is a need to run a job indefinitely following a schedule. These repetitive jobs can be utilized for recurring tasks such as backing up a database, pinging a server for health checks, creating snapshots of systems and much more.
```

docs/plus/cronjob.md
```diff
@@ -0,0 +1,93 @@
...
+* ### `startingDeadline`
+  Deadline for a job to run by if it misses its scheduled time. 
+  
+  For instance, if a job was scheduled to run every 5 minutes and you have mentioned a *startingDeadline* of 30 seconds. Then the job status would be updated as failed if the job does not start by 5 minutes and 30 seconds after the start time of previous job.
+
+* ### `successfulJobsRetained`
+  Number of successful job runs retained during CronJob execution.
+
+* ### `failedJobsRetained`
+  Number of failed job runs retained during CronJob execution.
```
iliapolo, line 21: This reads very much like an API reference. Remember the added value of a README is that it can be more use-case and intent focused. Instead of repeating the API reference, try formulating these capabilities in terms of use-cases, and provide code snippets for each. 

Also, put this after the basic usage example.

docs/plus/cronjob.md
```diff
@@ -0,0 +1,93 @@
...
+});
+
+// lets add a mysql container to our job. This adds this container to all pods created for this job.
+databaseBackupJob.addContainer({
+  image: 'databack/mysql-backup'
+});
+
+// lets setup a cron job here using the schedule method of Job construct.
+// schedule method creates a CronJob and returns a CronJob object.
+const backupCronJob = databaseBackupJob.schedule({
```
iliapolo, line 84: Not sure this actually makes sense. For the same reason we spoke of, because this eventually results in both a `Job` resource, and a `CronJob` resource, which isn't desirable. Lets drop this feature for now.

docs/plus/cronjob.md
```diff
@@ -0,0 +1,93 @@
...
+// lets define database backup job properties which we can pass to our CronJob resource
+const dbBackupJobProps: k8splus.JobProps = {
+  activeDeadline: Duration.minutes(5),
+  backoffLimit: 5,
+  containers: [
+    {
+      image: 'databack/mysql-backup',
+    }
+  ]
+};
```
iliapolo, line 47: Inline this into the `CronJob` instantiation, makes it easier to read and comprehend.

docs/plus/cronjob.md
```diff
@@ -0,0 +1,93 @@
...
+// lets setup a cron job here which backups mysql database.
+const backupCronJob = new k8splus.CronJob(this, 'BackupCronJob', {
+  jobProperties: dbBackupJobProps,
+  schedule: Schedule.rate(Duration.minutes(5)),
+  timezone: k8splus.Timezone.PDT,
+  concurrencyPolicy: k8splus.CronJobConcurrencyPolicy.ALLOW,
+  startingDeadline: Duration.seconds(30),
+  suspend: false,
+  successfulJobsRetained: 5,
+  failedJobsRetained: 5,
```
iliapolo, line 58: Don't include any non required properties. This should be the very most basic example.
Then other properties should be described in their own section along with an appropriate use case.

iliapolo (changes requested):

docs/plus/cronjob.md
vinayak-kukreja: Will add it.

docs/plus/cronjob.md
vinayak-kukreja: Will update it.

docs/plus/cronjob.md
vinayak-kukreja: Will update it.

docs/plus/cronjob.md
vinayak-kukreja: I see, makes sense, let me try to rephrase it.

docs/plus/cronjob.md
vinayak-kukreja: Right, will remove this.

docs/plus/cronjob.md
vinayak-kukreja: Will do.

docs/plus/cronjob.md
vinayak-kukreja: Ahh, ok. Will update it.

docs/plus/cronjob.md
```diff
@@ -0,0 +1,178 @@
...
+    backoffLimit: 5,
+    },
+  containers: [
+      {
+        image: 'databack/mysql-backup',
+      }
+    ]
+});
+```
+
```
iliapolo, line 37: This is where I would add a paragraph mentioning the important default values that we set:

*This will create a `CronJob` resource that runs the `databack/mysql-backup` every X minutes with an active deadline of 5 minutes and a backoff limit of 5*. To customize the properties, see [Link to API reference].

docs/plus/cronjob.md
```diff
@@ -0,0 +1,178 @@
...
+### You can also configure the parameters of a CronJob:
```
iliapolo, line 38: This entire section reads like an API reference. Focus on the important non trivial functionality here.

For example I would talk a little about the nice schedule helpers we provide (`Cron.daily()`, `Cron.hourly()`, ...).
I would mention any special validations we do, and why we do them.

iliapolo (changes requested):

iliapolo (approved):

Issue #883, opened 2022-03-15
Document aliased imports


iliapolo:
When importing two CRD's that share the same group id, one needs to use aliased imports, otherwise the imports will override one another.

> See https://github.com/prometheus-operator/kube-prometheus/tree/main/manifests/setup for an example of CRD's that share group id.

That is, instead of having:

```yaml
imports:
  - https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/main/manifests/setup/0alertmanagerConfigCustomResourceDefinition.yaml
  - https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/main/manifests/setup/0alertmanagerCustomResourceDefinition.yaml
```

We need to alias the imports like so:

```yaml
imports:
  - alertmanagerconfig:=https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/main/manifests/setup/0alertmanagerConfigCustomResourceDefinition.yaml
  - alertmanager:=https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/main/manifests/setup/0alertmanagerCustomResourceDefinition.yaml
```

This functionality is not documented. We should add to our CRD's [section](https://cdk8s.io/docs/latest/cli/import/#crds).

#### Side Note

It would be nice if the CLI could warn before overriding an import and suggest an alias import with a link to the new documentation section.

mcouthon:
Would it be possible to identify that both are coming from the same namespace, merge the two into a single eg Python file, and allow importing both from it?

iliapolo:
> @iliapolo do you think is this what happening there -> #881

I don't think it's related actually. I need to a closer look into what's happening there.

iliapolo:
@mcouthon 

> Would it be possible to identify that both are coming from the same namespace, merge the two into a single eg Python file, and allow importing both from it?

Should be doable I think. Can you create a feature request?

mcouthon:
Sure: #885 🙇

PR #1159, opened 2023-01-29
chore(docs): document aliased imports

ShubhikaBhardwaj:
### DESCRIPTION
Resolves https://github.com/cdk8s-team/cdk8s/issues/883
 
### CHANGES
Instead of importing like this:
```
imports:
 - https://raw.githubusercontent.com/jenkinsci/kubernetes-operator/6edd2e554f40d7ee03f1cf6e12feb278cb5097f0/deploy/crds/jenkins.io_jenkins_crd.yaml
````
Using alias for the imports:

```
imports:
  - io_jenkins_crd:=https://raw.githubusercontent.com/jenkinsci/kubernetes-operator/6edd2e554f40d7ee03f1cf6e12feb278cb5097f0/deploy/crds/jenkins.io_jenkins_crd.yaml
```

### DIFF
--- a/examples/go/crd/cdk8s.yaml
+++ b/examples/go/crd/cdk8s.yaml
@@ -2,7 +2,7 @@ language: go
 app: go run .
 imports:
   - k8s@1.17.0
-  - https://raw.githubusercontent.com/jenkinsci/kubernetes-operator/master/deploy/crds/jenkins.io_jenkins_crd.yaml
+  - io_jenkins_crd:=https://raw.githubusercontent.com/jenkinsci/kubernetes-operator/master/deploy/crds/jenkins.io_jenkins_crd.yaml
   - mattermost_crd.yaml
   - version.yaml
-  - https://github.com/knative/serving/releases/download/v0.14.0/serving-crds.yaml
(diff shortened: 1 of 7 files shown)


iliapolo (changes requested):
@ShubhikaBhardwaj Thanks for submitting this.

But the changes your are proposing is not exactly what the related issue is about. There's no need to use aliased imports for every single import, only for those that share a group id.
We prefer avoiding aliases when possible. 

So, the desired change is to actually just add documentation as to when aliased imports are needed to our [CRD's](https://cdk8s.io/docs/latest/cli/import/#crds) section on the website

ShubhikaBhardwaj:
Thanks for the clarification @iliapolo 
How do we identify a shared group id?

iliapolo:
@ShubhikaBhardwaj the group id is mentioned in the CRD itself, for example: https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/setup/0alertmanagerConfigCustomResourceDefinition.yaml#L9

So if one imports to different CRD's (e.g two different `Kind`s) that share the same group id, they need to use the alias.

Are you still able to pick this up?

PR #1191, opened 2023-02-18
fix(docs): use Node.of(self).id instead of Node.of(self).unique_id in getting-started python example

PriestOfAdanos:
While following python tutorial I noticed that instead of `id` it is using `unique_id` property, which seems to not be available or deprecated


iliapolo (approved):
Nice. Thanks!

Issue #1216, opened 2023-03-01
RFC | Resolve Cloud Tokens


iliapolo:
## Description

Allow cdk8s to resolve tokens from CDKs that provision cloud resources (e.g AWS CDK and CDKTF) during synthesis.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @iliapolo 
| Author(s)           | @iliapolo 
| API Bar Raiser      | @rix0rrr 
| Stakeholders        | @udaypant @evgenyka 

> See [RFC Process](https://github.com/cdk8s-team/cdk8s#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `rfc/proposed`)
- [x] API bar raiser assigned (ping us at [#cdk8s](https://cdk-dev.slack.com/archives/C0184GCBY4X) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `rfc/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `rfc/api-approved` applied to pull request)
- [x] Final comments period (label: `rfc/final-comments-period`)
- [x] Approved and merged (label: `rfc/approved`)
- [x] Execution plan submitted (label: `rfc/planning`)
- [x] Plan approved and merged (label: `rfc/implementing`)
- [ ] Implementation complete (label: `rfc/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

PR #1223, opened 2023-03-06
chore(rfc): resolve cloud tokens

iliapolo:
[Rendered Version](https://github.com/cdk8s-team/cdk8s/blob/epolon/rfc-resolve-cloud-tokens/rfc/1216-resolve-cloud-tokens.md)

Resolves https://github.com/cdk8s-team/cdk8s/issues/1216


examples/typescript/resolve-cloud-tokens/NOTES.md
```diff
@@ -0,0 +1,99 @@
...
+cdktf should download a state snapshot and store in cdktf.out so that automation running afterwards can consume the deployment information
+
+### AWS SDK V3
+
+- `const client = new aws.CloudControlClient()` should be possible since all config is optional.
+
+### AWS CDK
+
+- Why doesn't EIP logical id have a hash suffix?
+- Why don't logical ids contain the stack id?
```
rix0rrr, line 99: Because logical IDs are scoped to a stack, so they are calculated w.r.t. the containing Stack.

The nice property about this is that you can change the construct ID of your stack, or reparent it in the construct tree, without disturbing logical IDs. 

Everywhere we use `node.addr` (which uses the full construct path) is actually an antipattern, imo.

examples/typescript/resolve-cloud-tokens/NOTES.md
```diff
@@ -0,0 +1,99 @@
...
+
+cdktf should download a state snapshot and store in cdktf.out so that automation running afterwards can consume the deployment information
+
+### AWS SDK V3
+
+- `const client = new aws.CloudControlClient()` should be possible since all config is optional.
+
+### AWS CDK
+
+- Why doesn't EIP logical id have a hash suffix?
```
rix0rrr, line 98: All logical IDs of more than two construct ids, or more than 1 construct id where the 2nd one isn't `Resource` should have a hash suffix. Construct ids named `Default` are ignored from this calculation.

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+import * as kplus from 'cdk8s-plus-26';
+
+import { AwsCdkTokenResolver } from '@cdk8s/aws-cdk-token-resolver';
+
+const awsApp = new aws.App();
+const stack = new aws.Stack(awsApp, 'Stack');
+
+const resolver = new AwsCdkTokenResolver(stack);
+
+const k8sApp = new k8s.App();
```
rix0rrr, line 91: Wasn't the original vision to have everything in the same `App` ?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+    }
+ }]
+});
+
+awsApp.synth();
+k8sApp.synth();
+```
+
+Notice we create two applications: one for our cdk8s constructs, and one for our AWS CDK constructs.
+Both are defined and synthesized in the same file, but can be separated as needed.
```
rix0rrr, line 112: Not too far though, it will have to be the same application at runtime.

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+Since your Kubernetes resources now depend on AWS CDK deploy tokens, you'll first need to run `cdk deploy`, 
+and only then `cdk8s synth`.
+
+> Otherwise, the Kubernetes manifests will contain a string representation of the tokens (e.g `${Token[TOKEN.25]}`),
```
rix0rrr, line 116: How is it that if I don't run `cdk deploy`, my synthesized manifest will have an unsubstituted value?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+> Otherwise, the Kubernetes manifests will contain a string representation of the tokens (e.g `${Token[TOKEN.25]}`), 
+> instead of the concrete values. To learn more AWS CDK tokens, see [here](https://docs.aws.amazon.com/cdk/v2/guide/tokens.html).
+
+The implementation of the resolver utilizes [Cloud Control](https://docs.aws.amazon.com/cloudcontrolapi/latest/userguide/what-is-cloudcontrolapi.html)
+to fetch the runtime attributes of a resource. However, not all CloudFormation resources are currently supported by Cloud Control.
+
+> For a list of supported resources, see [here](https://docs.aws.amazon.com/cloudcontrolapi/latest/userguide/supported-resources.html).
+
+In case you need an attribute for such a resource, you can add a `CfnOutput` resource to your stack, and use the resolver to 
+explicitly fetch its value from the deployed stack. For example:
```
rix0rrr, line 125: This feels like typically the kind of thing the resolver should take care of transparently, no? 

If this is more reliable, why don't we pass everything via `CfnOutput` automatically?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+import * as aws from "@cdktf/provider-aws";
+import * as k8s from 'cdk8s';
+import * as kplus from 'cdk8s-plus-26';
+
+import { CdkTfTokenResolver } from '@cdk8s/cdktf-token-resolver';
+
+const awsApp = new tf.App();
+const stack = new tf.TerraformStack(awsApp, 'Stack');
+
+const resolver = new CdkTfTokenResolver(stack);
```
rix0rrr, line 167: How is the resolver checking the provenance of tokens? What if I want to use 2 resolvers at the same time (e.g., use values from AWS CDK and CDKTF both in my `Chart`?)

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+  }
+});
+```
+
+Your resolver will be invoked with the following arguments:
+
+- **context**
+  - *obj*: The `KubeService` instance of type `ApiObject`.
+  - *key*: `['spec', 'type']`
+- **value**: `LoadBalancer`
```
rix0rrr, line 67: Will the resolver literally get called for all values? Or just values that the systems *knows* are tokens?

rfc/1216-resolve-cloud-tokens.md
rix0rrr: Oh I see how this must work. The resolver is going to do an API call and if that call fails, it just won't substitute anything at all. Is that it?

Do you have a warnings mechanism, and in that case wouldn't it make more sense to return a `"<Fail: couldn't retrieve value>"` placeholder to make it crystal clear what went wrong?

rix0rrr: What do we do for permissions btw? No mention of this anymore.

Wouldn't using `CfnOutput`, combined with writing that output to a JSON file (`cdk deploy -o outputs.json`), be safer and faster?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+concrete value with a `fetchToken` function they implement:
+
+```ts
+...
+
+// define the bucket without an explicit name
+new aws.aws_s3.Bucket(stack, 'Bucket');
+
+// fetch its name directly from AWS
+const bucketName = fetchToken(bucket.bucketName);
```
rix0rrr, line 376: Unless Pulumi, this wouldn't work, would it? Because we'd need a deploy first as well?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+export class ResolutionContext {
+
+  public readonly obj: ApiObject;
+  public readonly key: string[];
+
+}
+
+export interface IResolver {
+
+  resolve(context: ResolutionContext, value: any): any;
```
rix0rrr, line 548: For future extensibility, I think I would offer an object with methods. Rather than:

```ts
if (shouldDoSomething(value)) {
  return doSomething(newValue); 
}
return value;
```

The implement would look like:

```ts
if (shouldDoSomething(value)) {
  context.replace(newValue); 
}
```

And future operations or tweaks could be made by extending the set of possible operations on `context`:

```ts
if (shouldDoSomething(value)) {
  context.replace(newValue, { cache: true }); 
}
```

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+```
+
+When a cdk8s chart is defined, a custom resolver can be passed to it:
+
+```ts
+new Chart(app, 'Chart', { resolver: new MyClassThatImplementsIResolver() })
+```
+
+During synthesis, the `resolve` function of the custom resolver will be invoked on every primitive value.
+The return value of this function will replace the original value and be written into the manifest.
```
rix0rrr, line 559: Do we do queries for every value on every synthesis? Isn't that a bit wasteful?

rix0rrr: Also this breaks network-isolated synthesis -- is there a way we can write instructions somewhere to "query later" ?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+    E("Extract Physical Resource ID")-->|"Attribute !== 'Ref'"|G("cloudcontrol.GetResource")
+    G("cloudcontrol.GetResource")-->H("Extract Resource Properties")
+    H("Extract Resource Properties")-->I("Return Attribute Value")
+```
+
+> Note that the PoC code doesn't currently contain the `fetchOutput` method described in the README.
+> We assume this function is fairly easy to implement by invoking `cloudformation.DescribeStacks` and extracting
+> the output value.
+
+**There are some noteworthy points to pay attention to:**
```
rix0rrr, line 587: All the more reason to go via `CfnOutput` by default?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+```
+
+This solution has the benefit of automatically supporting every value that 
+can be defined in an AWS CDK application. It also simplifies the fetching logic because 
+it requires a single `cloudformation.DescribeStacks` call. 
+
+However, this option was discarded because of usability concerns:
+
+- Requires synthesis of the cdk8s app to happen before the AWS CDK app. This can only be 
+controlled by the user, and therefore error prone.
```
rix0rrr, line 650: Error prone is not really the argument. You can throw errors if the preconditions aren't met (`Error: CDK app already synthesized. Call 'cdk8sapp.synthesize()` first.`), which makes it impossible to get wrong.

You could say it's poor user experience. Maybe.

Wouldn't putting all constructs in the same `App` solve that?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+- Synthesizing the AWS CDK application separately from the cdk8s application will result in a different cloud assembly.
```
rix0rrr, line 651: I'm not following this.

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+- The cdk8s application needs to be synthesized twice, once to add the CloudFormation outputs, and once to fetch their values.
```
rix0rrr, line 652: Yes, but right now saying "do `cdk deploy` first and then `cdk8s synth`" is effectively the same thing.

It's *also* synthesizing twice.

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+it requires a single `cloudformation.DescribeStacks` call. 
+
+However, this option was discarded because of usability concerns:
+
+- Requires synthesis of the cdk8s app to happen before the AWS CDK app. This can only be 
+controlled by the user, and therefore error prone.
+- Synthesizing the AWS CDK application separately from the cdk8s application will result in a different cloud assembly.
+- The cdk8s application needs to be synthesized twice, once to add the CloudFormation outputs, and once to fetch their values. 
+This means that the first synthesis will inherently produce an invalid manifest.
+- Will create a cyclic dependency in case the cdk8s and AWS CDK application are in different repositories / packages.
```
rix0rrr, line 654: I don't get this concern either. Why? The CDK app doesn't need to know about the identity of its consumer, so looks to me the dependency arrow only points one way?

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+    thing worth mentioning though is that the AWS CDK logical IDs [**do not contain the 
+    id of the stack**](https://github.com/aws/aws-cdk/blob/main/packages/aws-cdk-lib/core/lib/stack.ts#L1273), 
+    contradictory to what the documentation](https://docs.aws.amazon.com/cdk/v2/guide/identifiers.html#identifiers_logical_ids) states:
+
+    > *For example, the Amazon S3 bucket in the previous example that is created within 
+    > `Stack2` results in an `AWS::S3::Bucket` resource. The resource's logical ID 
+    > is `Stack2MyBucket4DD88B4F` in the resulting AWS CloudFormation template.*
+
+    This makes it impossible to automatically detect which stack a resource belongs to 
+    based solely on its logical ID.
```
rix0rrr, line 749: Don't use logical ID. 

But remember that CDK must know what stack a value is produced by already, otherwise automatic cross-stack references wouldn't work.

A `CfnReference` is a type of token that has a `target` (or `anchor`, I forget the name), which is the `CfnElement` that produces the value. You can `Stack.of(reference.target)` to get the producing stack.

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+
+## Appendix
+
+### `Ref` vs `PhysicalResourceID`
```
rix0rrr, line 753: All the concerns around CC-API and the identifiers make me not like this option.

rfc/1216-resolve-cloud-tokens.md
```diff
@@ -0,0 +1,845 @@
...
+[ ] Signed-off by API Bar Raiser @xxxxx
+```
+
+## Public FAQ
+
+### What are we launching today?
+
+This launch consists of three deliverables:
+
+- A new feature in the `cdk8s` core library that allows injecting custom resolution logic during synthesis.
```
rix0rrr, line 213: Given that now all 3 CDKs have tokens or token-related features... wouldn't it make sense to put down some motivation why this feature shouldn't be added to `constructs` by now? 

Yes, it would probably become a bigger project. But wouldn't it be better for everyone to rethink this infrastructure from the ground up, including all the features we're now stitching on to it?

rfc/1216-resolve-cloud-tokens.md
iliapolo: Not really. As we talked offline, the only single `App` that makes sense is a sort of "constructs" App that can synthesize two different artifacts, which we don't have. But even that doesn't really fit this use case because conceptually this is about passing information from one app to another, as they each have a separate synthesis and deployment lifecycle.

rfc/1216-resolve-cloud-tokens.md
iliapolo: Revised

rfc/1216-resolve-cloud-tokens.md
iliapolo: As we talked about, yes decided to go with `CfnOutput` eventually. As for the premature synthesis, I decided to fail it in case the resources aren't available - because the manifest is not deployable. This way we can throw a descriptive error notifying the user that they probably didn't deploy the CDK application yet.

rfc/1216-resolve-cloud-tokens.md
iliapolo: Each resolver uses facilities provided by the CDK framework to check (i.e `Token.isUnresolved`). 

Currently the implementation only allows passing one resolver to the cdk8s app, so if you actually have values from both CDKTF and AWS CDK, e.g:

```ts
new ApiObject(this, 'obj', {
  spec: {
    prop1: <token-from-aws-cdk>,
    prop2: <token-from-cdktf>
  }
});
```

You would have to implement your own resolver that delegates to the specific ones. I suppose we could vend such a resolver ourselves if we see it is needed. But I doubt there are many users operating both AWS CDK and CDKTF...

Another interesting use case is:

```ts
new ApiObject(this, 'obj', {
  spec: {
    prop1: <token-from-aws-cdk> + <token-from-cdktf>,
  }
});
```

I haven't considered this, not even sure this is possible...but again I doubt this is an actual use case.
Interested in your thoughts here though.

rfc/1216-resolve-cloud-tokens.md
iliapolo: The _system_ doesn't know anything about tokens, so yes the resolver will get called for all values. The resolver must check that it is being passed a token, and no-op on literal values.

rfc/1216-resolve-cloud-tokens.md
iliapolo: Done 👍

rfc/1216-resolve-cloud-tokens.md
iliapolo: > Do we do queries for every value on every synthesis? Isn't that a bit wasteful?

Possibly. Are you thinking about introducing a context file? I'm not keen on doing that before we see it actually poses a problem but interested what you think.

> Also this breaks network-isolated synthesis -- is there a way we can write instructions somewhere to "query later" ?

In cdk8s there is only `synth`, no `deploy`. So there is no "later"...yes this means that isolated environment won't be able to synthesize such applications. This does make me think that maybe we shouldn't fail synthesis in case we can't perform the query, because it may also have a bad affect on local devs executing synth - they might not have the necessary credentials, or may not even need that part of the manifest for what they are currently developing an testing.

rfc/1216-resolve-cloud-tokens.md
iliapolo: Removed this concern and focused solely on the circular dependency and awkward deployment workflow.

rfc/1216-resolve-cloud-tokens.md
iliapolo: It was to say that its weird for a CDK application to produce a different cfn template based on whether or not the application is being consumed by cdk8s or not. But I agree its not a strong argument, removed it.

rfc/1216-resolve-cloud-tokens.md
iliapolo: Removed

rfc/1216-resolve-cloud-tokens.md
iliapolo: Dropped :)

rfc/1216-resolve-cloud-tokens.md
iliapolo: As we discussed, this feature is not introducing a new token system into cdk8s, it merely allows interpreting tokens of other systems. And since we are using `CfnOutput` instead of Cloud Control, we don't even need to replicate the handling of CloudFormation intrinsics, so I don't see anything thats worth pulling out to `constructs`.

rfc/1216-resolve-cloud-tokens.md
rix0rrr: I wonder if it might make sense to make this a 2-step process.

The framework/resolver writes instructions on how to fetch the value to a file.

When can then choose to have:

```
# Do everything
$ cdk8s synth 

# Do it in 2 steps, second one runs with necessary credentials, potentially later
$ cdk8s synth --no-fetch

$ export AWS_ACCESS_KEY_ID=****
$ cdk8s fetch-values
```

rfc/1216-resolve-cloud-tokens.md
rix0rrr: Given the resolver API we currently have, which I really like, I think we could trivially allow the first case by allowing this:

```ts
const k8sApp = new k8s.App({ 
  resolvers: [new AwsCdkResolver(), new CdkTfResolver()],
});
```

Iterating over all resolvers, and stopping as soon as one of them calls `context.replaceValue()`.

---------

The second case, where we have `token1 + token2` can only work if we can peel apart the separate tokens, and that can only work if the token formatting is standardized, which it currently is not.

So *shrug*, might be useful but can't be in scope for this.

iliapolo: Yeah you're right about the first case - I'll make the change. Regarding the second one, if we see a demand for it, it might be a good forcing function to standardize tokens across the board and put them in `constructs` with a consistent formatting.

rfc/1216-resolve-cloud-tokens.md
iliapolo: I would use `--no-resolve` and `cdk8s resolve` as the semantic, because not all resolvers necessarily "fetch" values. Let me think about this.

PR #1225, opened 2023-03-06
fix(docs): documentation about resource name generation

### LINKED ISSUE #232: Documentation about name generation

Names (as in `.metadata.name`) are automatically generated from the name given to the construct. This is a request to add to the documentation: How are those names generated, especially how stable are they? (As in when do they change)

As #217 points out is it possible to workaround this but I like to add the reasoning why this is done to the documentation and what use cases this supports. And maybe even add the workaround for use cases where this might be harmful.

---

sumupitchayan:
Fixes #232 which requests additional documentation regarding resource name generation in cdk8s.


docs/basics/api-object.md
```diff
@@ -10,3 +10,50 @@ It is possible to customize the prefix by providing a string value to the
 `--class-prefix` flag, or by passing `--no-class-prefix`. (Note: if no prefix is
 used for k8s resources, this may lead to conflicts - see
 https://github.com/cdk8s-team/cdk8s/issues/140).
+
+### Resource Names
+
+Every object has a name that is autogenerated when the cdk8s app is synthesized.
```
vinayak-kukreja, line 16: Instead of `object` this can be `resource`.

docs/basics/api-object.md
```diff
@@ -10,3 +10,50 @@ It is possible to customize the prefix by providing a string value to the
...
+    disableResourceNameHashes: true,
+});
+
+new kplus.Deployment(chart, 'deployment', {
+  containers: [{ image: 'nginx' }]
+});
+
+app.synth();
+```
+Which would similarly generate a `Deployment` now with the name: `my-chart-deployment`.
```
vinayak-kukreja, line 59: Have we mentioned anywhere what issues can non unique identifier cause for the customer?

vinayak-kukreja (changes requested):

docs/basics/api-object.md
sumupitchayan: Added a line about how unique names are required in K8s - let me know if this is sufficient

docs/basics/api-object.md
```diff
@@ -10,3 +10,51 @@ It is possible to customize the prefix by providing a string value to the
 `--class-prefix` flag, or by passing `--no-class-prefix`. (Note: if no prefix is
 used for k8s resources, this may lead to conflicts - see
 https://github.com/cdk8s-team/cdk8s/issues/140).
+
+### Resource Names
+
+Every resource has a name that is autogenerated when the cdk8s app is synthesized.
+The name is the passed in string you assign to an object, prefixed by the name of the
+chart it is contained in, and suffixed by the hash of the construct path.
```
iliapolo, line 18:
```suggestion
By default, cdk8s will autogenerate a name for every resource when it is instantiated.
The name is created from the path of the construct, and suffixed with a hash of that path.
```

docs/basics/api-object.md
```diff
@@ -10,3 +10,51 @@ It is possible to customize the prefix by providing a string value to the
...
+const chart = new MyChart(app, 'my-chart');
+
+new kplus.Deployment(chart, 'deployment', {
+  containers: [{ image: 'nginx' }]
+});
+
+app.synth();
+```
+
+Generates a `Deployment` with the name: `my-chart-deployment-c8c354dd`.
```
iliapolo, line 36:
```suggestion
Generates a `Deployment` with the name: `my-chart-deployment-c8c354dd`. Where `my-chart` is the id of the `Chart`, `deployment` is the id of the `Deployment`, and `c8c354dd` is the hash suffix.
```

docs/basics/api-object.md
```diff
@@ -10,3 +10,51 @@ It is possible to customize the prefix by providing a string value to the
...
+
+In Kubernetes, only one resource of a given kind can have a given name at a time.
+Hashes are used to ensure that all resources have unique identifiers. However, they are
+not required and it may sometimes be undesirable to include the hash suffixes.
```
iliapolo, line 40:
```suggestion
You can programmatically access the name of any resource in a cdk8s app by accessing the `.name` property of a resource. For example:

```ts
const deployment = new kplus.Deployment(chart, 'deployment', {
  containers: [{ image: 'nginx' }]
});
const deploymentName = deployment.name;
```
```

File: docs/basics/api-object.md [hunk 6]
```diff
@@ -10,3 +10,51 @@ It is possible to customize the prefix by providing a string value to the
...
+
+To remove hashes from resource names, you can change the above code by setting the
+`ChartProps.disableResourceNameHashes` value to `true` like so:
```
iliapolo, line 43:
```suggestion
To remove hashes from resource names, you can set the `disableResourceNameHashes` property to `true`:
```

docs/basics/api-object.md
```diff
@@ -10,3 +10,51 @@ It is possible to customize the prefix by providing a string value to the
...
+const app = new App();
+const chart = new MyChart(app, 'my-chart', {
+    disableResourceNameHashes: true,
+});
+
+new kplus.Deployment(chart, 'deployment', {
+  containers: [{ image: 'nginx' }]
+});
+
+app.synth();
```
iliapolo, line 58:
```suggestion
...
const chart = new MyChart(app, 'my-chart', {
    disableResourceNameHashes: true,
});
...
```

docs/basics/api-object.md
```diff
@@ -10,3 +10,51 @@ It is possible to customize the prefix by providing a string value to the
...
+```
+Which would similarly generate a `Deployment` now with the name: `my-chart-deployment`.
```
iliapolo, line 60:
```suggestion
Which would similarly generate a `Deployment` with the name: `my-chart-deployment`. Note that disabling hashes introduces the possibility of name clashes. For example, the following code:

```ts
// think of this as a custom construct
const c = new Construct(this, 'Construct');

new kplus.Deployment(c, 'Deployment', {
  containers: [{ image: 'image' }]
});

new kplus.Deployment(this, 'Construct-Deployment', {
  containers: [{ image: 'image' }]
})
```

Will produce two deployments with the same name: `chart-construct-deployment`. This happens because cdk8s uses a hyphen (`-`) as the delimiter between elements in the path, so if the id of your construct also contains a hyphen (`-`), a collision can occur.

In addition, you can explicitly set a name for every resource by using the `metadata.name` property:

```ts
new kplus.Deployment(c, 'Deployment', {
  metadata: {
    name: 'my-deployment'
  },
  ...
});
```

```

iliapolo (changes requested):

iliapolo (approved):

PR #1227, opened 2023-03-07
Update ADOPTERS.md

Vad1mo:
We are using cdk8s and cdk8s+ and are super happy users.

**Cdk8s is the best thing, since the invention of YAML to describe Kubernetes Manifests.**


iliapolo (approved):
Sweet, thanks @Vad1mo!

PR #1233, opened 2023-03-07
chore(examples): add nextcloud

SmsS4:
Hello i wrote an example for nextcloud + grafana using cdk8s in typescript.
This example includes:
- nextcloud
- mariadb
- redis
- traefik
- cert-manager
- certificate issuer
- victoria metrics
- grafana
and these components of cdk8s and cdk8s+
- Helm
- Secret
- Deployment
- Container
- EnvValue
- Service
- Ingress
- PVC
- ApiObject and JsonPatch
- ConfigMap
- InitContainer

I had to read lots of cdk8s's document and other examples to write this, so i think this example is not that bad :))

### DIFF
--- a/examples/nextcloud-with-grafana/README.md
+++ b/examples/nextcloud-with-grafana/README.md
@@ -1,0 +1,34 @@
+This example creates
+- nextcloud
+- mariadb
+- redis
+- traefik
+- cert-manager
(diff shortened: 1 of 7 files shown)


SmsS4:
Opened PR in cdk8s-examples

PR #1263, opened 2023-03-30
chore(adopters): https://griffin.com

dancmeyers:
@iliapolo asked if we would be OK being on the adopters list. I've run it by our CTO, and it's a 👍, so here's the PR.

I also fixed the alphabetisation and made a slight grammar change. That's a separate commit so can be easily binned if desired.


iliapolo:
@dancmeyers Do I have your permission to set DCO to pass?

iliapolo (approved):

Issue #1271, opened 2023-04-05
RFC | Helm Import


vinayak-kukreja:
## Description

Currently cdk8s supports adding helm charts as construct with the [helm construct](https://cdk8s.io/docs/latest/basics/helm/). This approach does not take into consideration data types for the values that the helm chart can accept.

This RFC is proposing to enable users to import a helm chart as a construct using `cdk8s import`. This would auto generate a construct for the user and if a `values.schema.json` is present in the helm chart, then the `values` properties would have appropriate types associated with them.

## Roles

| Role                | User
|---------------------|------------------------------
| Proposed by         | @vinayak-kukreja 
| Author(s)           | @vinayak-kukreja 
| API Bar Raiser      | @iliapolo 
| Stakeholders        | @cdk8s-team/cdk8s-team 

> See [RFC Process](https://github.com/cdk8s-team/cdk8s#rfc-process) for details

## Workflow

- [x] Tracking issue created (label: `rfc/proposed`)
- [x] API bar raiser assigned (ping us at [#cdk8s](https://cdk-dev.slack.com/archives/C0184GCBY4X) if needed)
- [x] Kick off meeting
- [x] RFC pull request submitted (label: `rfc/review`)
- [x] Community reach out (via Slack and/or Twitter)
- [x] API signed-off (label `rfc/api-approved` applied to pull request)
- [x] Final comments period (label: `rfc/final-comments-period`)
- [x] Approved and merged (label: `rfc/approved`)
- [x] Execution plan submitted (label: `rfc/planning`)
- [x] Plan approved and merged (label: `rfc/implementing`)
- [ ] Implementation complete (label: `rfc/done`)

---

> Author is responsible to progress the RFC according to this checklist, and
apply the relevant labels to this issue so that the RFC table in README gets
updated.

yuriy-yarosh:
I've made a basic helm charts import with codegen over imported k8s structs, using prerendered `helm template` charts, for golang. The issue I've faced is that it required a stacked codegen to reduce the boilerplate for golang specifically. I could contribute the respective JSII bindings as well, but the user still would need to provide a `values.yaml` for the `helm template` somewhere.

iliapolo:
@yuriy-yarosh the values will still be provided as normal. But instead of providing it to a generic `Helm` construct, they will provide it to a specific `Airflow` construct (for example), where the values dictionary will now be typed. 

Im not sure I understand what you mean by stacked codegen, do you have an example to show?

PR #1272, opened 2023-04-05
chore(rfc): helm import

vinayak-kukreja:
[Tracking Issue](https://github.com/cdk8s-team/cdk8s/issues/1271)


rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+new MyChart(app, 'typescript-demo');
+app.synth();
+```
+
+> **Note:**
+>
+> * You would need `Helm` to be installed on your machine for using this feature. 
+> * We assume that you are authenticated to access the helm chart being mentioned in the url.
+> * If a helm chart contains a `values.schema.json` then the `values` properties within the construct properties for the chart would have specific types. For instance, in the prior example `fullnameOverride` is not of `any` type and instead is of `string` type since `airflow` chart has a `values.schema.json` file in it.
+
```
iliapolo, line 50: This should be front and center - not as a note. Also lets make sure the example chart we use actually does have a `values.schema.json` file and mention that if *your* chart doesn't, the values won't be typed.

rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+}
+
+const app = new App();
+new MyChart(app, 'typescript-demo');
+app.synth();
+```
+
+> **Note:**
+>
+> * You would need `Helm` to be installed on your machine for using this feature.
```
iliapolo, line 47:
```suggestion
> * You would need `helm` to be installed on your machine for using this feature. 
```

rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+
+const app = new App();
+new MyChart(app, 'typescript-demo');
+app.synth();
+```
+
+> **Note:**
+>
+> * You would need `Helm` to be installed on your machine for using this feature. 
+> * We assume that you are authenticated to access the helm chart being mentioned in the url.
```
iliapolo, line 48:
```suggestion
> * For accessing private helm repositories, you must be authenticated to the repository in a way that the `helm pull` command recognizes.
```

iliapolo: We should also add this disclaimer to our current Helm documentation :)

rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+>       fs.rmSync(tmpDir, { recursive: true });
+>    }
+>   ```
+
+#### Code Generation
+
+ImportBase class has abstract methods [moduleNames](https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/src/import/base.ts#L40) and [generateTypeScript](https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/src/import/base.ts#L42). As part of the new class `ImportHelm` we also need to define these. Because, when the [import](https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/src/import/base.ts#L44) function is [invoked]((https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/src/import/dispatch.ts#L17)), it invokes the [generateTypeScript](https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/src/import/base.ts#L80-L82) function. 
+
+```typescript
+  public get moduleNames() {
```
iliapolo, line 243: Can you explain what this method is used for and why you decided you return the chart name? is this the name of the generated `.ts` file?

rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+      }
+    });
+  }
+}
+
+const app = new App();
+new MyChart(app, 'typescript-demo');
+app.synth();
+```
+
```
iliapolo, line 44: Lets also add how the generated `airflow.ts` file will look like. This would make it clear that the values are typed.

rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+As mentioned above, we do currently have a [solution]((https://cdk8s.io/docs/latest/basics/helm/)) for including helm charts in cdk8s app. The proposed feature is definitely an enhancement but would take up developers time and effort to implement and maintain.
+
+### What is the technical solution (design) of this feature?
+
+#### Helm chart url
+
+The following would be the format for a helm url being passed into the import command:
+
+```
+cdk8s import helm:<repo-url>/<chart-name>@<chart-version>
```
iliapolo, line 90: Are you assuming that the chart name is the last element separated by a `/`? Is this always true? Is this what `helm pull` assumes?

rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+    code.line();
+
+    emitConstruct();                                            // <--------------- Creating construct for helm chart
+
+    function emitPropsInterface() {
+      code.openBlock(`export interface ${chartName}Props`);
+
+      code.line('readonly chart?: string;');
+      code.line('readonly repo?: string;');
+      code.line('readonly version?: string;');
```
iliapolo, line 326: These shouldn't be part of the interface though. We should just set them to the value (that we already know) and use them in the implementation.

rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+
+### What are the drawbacks of this solution?
+
+* **Interface similar to HelmProps**
+  
+  As mentioned in the previous sections, there is an interface similar to HelmProps in the proposed solution due to a jsii limitation. We can use `mgrain's` [project](https://github.com/mrgrain/jsii-struct-builder) to reduce code but the emitted interface would still be similar to HelmProps. 
+
+* **Multiple values.schema.json**
+
+  We are considering only the `values.schema.json` present at the root of the chart since helm by default looks for `values.yaml` at the root of the chart. I have seen multiple values.yaml files in a chart's sub-folders and these can have schema associated with them.
```
iliapolo, line 482: Can you give an example?

rfc/helm-import.md
```diff
@@ -0,0 +1,515 @@
...
+
+  Building on previous point, if we just include one schema file for code generation, it can lead to missing properties that the user is able to provision for manually when using `values.yaml` file. 
+
+  To resolve this, we can add an `additionalProperties` property to `values` which can get any additional property a user wants to add other than the auto generated properties.
+
+### What is the high-level project plan?
+
+* Addressing the open issues.
+* Implementing the feature and adding tests for it. 
+* Initially this feature can be marked as experimental for the users. In this time we can keep track of issues being created for the feature and if there are feature requests for it.
```
iliapolo, line 496: Feels a little too narrow to be treated as experimental. I think can just go stable from the start.

rfc/helm-import.md
vinayak-kukreja: Makes sense, will add it.

rfc/helm-import.md
vinayak-kukreja: Good callout, on it. :)

rfc/helm-import.md
vinayak-kukreja: I see. `airflow` does have a schema file. Are you suggesting that we emit a warning to the user if the chart they are using does not have a schema file and let them know that the value property would not be typed?

rfc/helm-import.md
vinayak-kukreja: Yes, I used it here since this would be used for name of the generated file. Although, now I see that `module` does not sound correct for this.

vinayak-kukreja: Link: https://github.com/cdk8s-team/cdk8s-cli/blob/2.x/src/import/base.ts#L40

rfc/helm-import.md
vinayak-kukreja: Ahh, correct. Will change it.

rfc/helm-import.md
vinayak-kukreja: I was not able to find the chart that I encountered this in, will keep looking. 

But, I found a [GH issue](https://github.com/helm/helm/issues/11815) for requesting support for multiple schema files and that was not entertained. So I probably misunderstood.

rfc/helm-import.md
iliapolo: No I don't mean with a warning - just in the documentation.

rfc/helm-import.md
iliapolo: Instead of assuming it. I would just treat whatever is between `helm:` and `@` as the "thing" you pass on to `helm pull` - then to get the chart name, just look for the Chart.yaml file in the output of `helm pull`.

rfc/helm-import.md
iliapolo: I think you might have seen sub-charts. However, this does bring up the question of what to do with sub charts? 

If a chart has sub-charts, we need to make sure those sub-charts can be configured as well. And if the sub-chart has a `values.schema.json` - that configuration should be typed as well. Lets talk about this online.

rfc/helm-import.md
vinayak-kukreja: Hey, I think this `helm:<repo-url>/<chart-name>@<chart-version>` should be our syntax for providing a url. The reason for this being, 
* `repo-url` would be the chart repository url, like `https://charts.bitnami.com/bitnami` and this in `helm pull` is passed in with the `--repo` flag.
* `chart-name` is used to get a particular chart from the repo. 
For instance,
```
helm pull mysql --repo https://charts.bitnami.com/bitnami/
```

------

`helm pull` either accepts `absolute URLs(https://charts.bitnami.com/bitnami/mysql-9.10.10.tgz)` or `repo/chartname`(which is what we have after adding the repo like [this](https://artifacthub.io/packages/helm/bitnami/mysql?modal=install)). 

I can run the `helm repo add` for adding the chart and then pulling it, but I was avoiding it so that `helm pull` does not change anything locally for the user. 


-----

Do you think `helm:<repo-url>/<chart-name>@<chart-version>` could be restricting since we do not know what urls are possible?

vinayak-kukreja: Ohh, I see we do callout, `helm repo add` with our current helm implementation. We can do the same for this. WDYT?

Link: https://github.com/cdk8s-team/cdk8s/blob/master/docs/basics/helm.md?plain=1#L14

rfc/helm-import.md
vinayak-kukreja: PR: https://github.com/cdk8s-team/cdk8s/pull/1411

rfc/helm-import.md
vinayak-kukreja: Synced offline and we are moving forward with `helm:<repo-url>/<chart-name>@<chart-version>` syntax.

iliapolo (approved):

PR #1484, opened 2023-09-09
chore(adopters): https://canida.io

2start:
Hello,

I'd like to contribute to cdk8s. The first step that was proposed was to extend the list of adopters.
I think cdk8s is awesome because I see the following shortcomings with other approaches:

kustomize: Get's really complicated really fast. Requires a lot of search & replace when modifying manifests.
Helm: Easy to package & deploy independent services. However, it's hard to manage the interdependencies of multiple services. 
In general: Hard to abstract away commonly used resources. E.g. in cdk8s I got a folder `custom-resources` which allows me to easily deploy preconfigured resources such as a new Ingress based on my custom load balancer controller. No need to copy around the configuration X times or look into 3 layers of kustomize abstractions. Just a simple class that can be modified with a handful of parameters e.g. if the ingress should be internal / public.


iliapolo (approved):
@2start Awesome! Happy to hear you are using cdk8s. And contributions of any kind are more than welcome :)

PR #1538, opened 2023-10-03
chore(website): resolver docs touchups

iliapolo:
- Resolver interface name was wrong. Left over from its previous name
- Replace the example resolver with a slightly more realistic one
- Remove explanation about invocation patterns, its kind of overwhelming and un-necessary now that the example contains a real implementation

### DIFF
--- a/docs/basics/app.md
+++ b/docs/basics/app.md
@@ -34,17 +34,31 @@ app.synth();
 ## Resolvers
 
 Resolvers are a mechanism to inject custom logic into the cdk8s value resolution process. 
-It allows to transform any value just before being written to the Kubernetes manifest. To define a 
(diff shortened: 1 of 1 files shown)


docs/basics/app.md
```diff
@@ -34,17 +34,31 @@ app.synth();
...
+It allows to transform any value just before being written to the Kubernetes manifest. 
+
+For example, lets create a resolver that adds a prefix to every resource in our application.
+To define a custom resolver, create a class that implements the `IResolver` interface:
 
 ```ts
 import { IResolver, ResolutionContext } from 'cdk8s';
 
-export class MyCustomResolver implements IValueResolver {
+export class AddNamePrefixResolver implements IResolver {
```
vinayak-kukreja, line 45: Love the example :D

PR #1428, opened 2023-08-15
chore: automate kubernetes upgrades

sumupitchayan:
Closes [#2455](https://github.com/cdk8s-team/cdk8s-plus/issues/2455)

When a new version of Kubernetes is released every quarter, we need to upgrade the CDK8s codebase by importing the new spec, creating a new branch in the cdk8s-plus repo and updating the website/documentation. This is currently done via a manual process (outlined [here](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-26/main/upgrade-k8s-version.md)) that takes up a decent chunk of engineering time and is prone to human errors. This project creates an automation workflow to prevent the manual work of this process for future releases of Kubernetes.

There are two pre-requisite PR’s that have to be merged in other repos before this PR:

* https://github.com/cdk8s-team/cdk8s-plus/pull/2702
* https://github.com/cdklabs/cdk-ops/pull/2852

PR #1574, opened 2023-10-18
fix(workflows): auto-approve labelling conditional statement always runs true

sumupitchayan:
Follow up to #1428 

We initially wanted to conditionally set the `auto-approve` label of PR's in this workflow based on a `testingMode` input parameter. This first step is supposed to conditionally set the `labels` output to `auto-approve` if `testingMode` was set to false:

```ts
{
  name: 'Set auto-approve label for PR if in testing mode',
  id: 'set-auto-approve-label',
  if: '${{ github.event.inputs.testingMode }} == false',
  run: 'echo labels="auto-approve" >> $GITHUB_OUTPUT',
  env: { GITHUB_TOKEN: '${{ secrets.PROJEN_GITHUB_TOKEN }}' },
  continueOnError: false,
},
```

And would then be used in the following step creating a PR like so:
```
...WorkflowActions.createPullRequest({
  ... ,
  labels: [
    '${{ steps.set-auto-approve-label.outputs.labels }}',
  ],
}),
```

However, there is a very annoying [known bug in GitHub workflow steps](https://github.com/actions/runner/issues/1173) where conditions are always evaluated as true when containing expression syntax inside. This causes the `auto-approve` label to always be added to PR's in the Kubernetes upgrade workflow, even when `testingMode` is set to `true`.

I've had a hard time finding workarounds to this problem. Right now what I'm just doing is manually commenting-out the steps that involve publishing or setting the `auto-approve` label of PR's created.


iliapolo:
@sumupitchayan Can't you move the condition out of the `if` and inside the `run` directive. You would then set the step output to "auto-approve" if the condition is true, and to an empty string (or whatever can work that causes a no-op) when the condition is false.

sumupitchayan:
> @sumupitchayan Can't you move the condition out of the `if` and inside the `run` directive. You would then set the step output to "auto-approve" if the condition is true, and to an empty string (or whatever can work that causes a no-op) when the condition is false.

@iliapolo I tried doing that earlier with this:

```ts
        {
          name: 'Set auto-approve label for PR if in testing mode',
          id: 'set-auto-approve-label',
          run: 'if (${{ github.event.inputs.testingMode }} == false);then echo labels="auto-approve" >> $GITHUB_OUTPUT;fi',
          env: { GITHUB_TOKEN: '${{ secrets.PROJEN_GITHUB_TOKEN }}' },
          continueOnError: false,
        },
        ...WorkflowActions.createPullRequest({
          ...
          labels: [
            '${{ steps.set-auto-approve-label.outputs.labels }}',
          ],
        }),
```

But it doesn't seem to be working in the `Create PR` step of the `generate-new-k8s-spec` job: [see the logs here](https://github.com/cdk8s-team/cdk8s/actions/runs/6576079245/job/17864657727) -- no `auto-approve` label is set on the PR even when I set the `testingMode` default to `false`

Issue #2132, opened 2024-09-16
Roadmap link in the readme is broken


renannprado:
<img width="636" alt="image" src="https://github.com/user-attachments/assets/fba0af0c-fab4-49f8-8530-77ad764d8216">

parthlaw:
What should be the actual link? The url points to projects but there is no project in the Projects section.

applike-ss:
@mrgrain Is it possible the project is setup to private instead of publich in terms of access permissions?

mrgrain:
> @mrgrain Is it possible the project is setup to private instead of publich in terms of access permissions?

No, I don't see any projects either.

renannprado:
@mrgrain @parthlaw @applike-ss if this is resolved, maybe it also helps to clean this doubt here:

https://github.com/cdk8s-team/cdk8s/discussions/2052 Is this project still being maintained?

applike-ss:
To me it is not resolved and the project status is unknown

PR #2286, opened 2024-12-09
chore: clarify that roadmap is still being discussed

iliapolo:
Resolves https://github.com/cdk8s-team/cdk8s/issues/2132. 

Not exactly a resolution to the general roadmap discussion but at least the link won't be broken.


saiyush (approved):

saiyush (approved):
approved

PR #2477, opened 2025-04-10
feat: add docs type for conventional commits

### LINKED ISSUE #2476: [Meta] Add `docs` type for convential commits.
Add `docs` as an acceptable type for [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), this repo only has the docs for cdk8s at this point, so changes to the docs aren't really a feature or fix to the underlying library.

### Use Case:
Adding to the existing documentation, fixing typos, etc.

### Proposed Solution:
Change the config for the workflow that checks the types.

---

clayrosenthal:
Fixes #2476, allows `docs:` type for PRs

### DIFF
--- a/.github/workflows/pull-request-lint.yml
+++ b/.github/workflows/pull-request-lint.yml
@@ -27,4 +27,5 @@ jobs:
             feat
             fix
             chore
+            docs
           requireScope: false


iliapolo (changes requested):
@clayrosenthal appreciate the PR, but this needs to be done via `projen` (note that this file generated). We use a common [repository](https://github.com/cdk8s-team/cdk8s-projen-common) for all our projen based config, so it needs to be changed over there and it will then be automatically applied to all our repos.

Let me know if you want to pursue it and I can provide more guidance.

clayrosenthal:
Sure I can take a shot at that. Would I change the same `.github/` file in that repo or something else in there?

clayrosenthal:
Opened a PR in the [projen repo](https://github.com/cdk8s-team/cdk8s-projen-common/pull/1255)

PR #2706, opened 2025-09-17
chore(docs): fix possessive typo in Helm support docs

johngibb:
This patch corrects "it's" → "its" in the Helm documentation.

The word is being used as a possessive pronoun, not a contraction,
so the apostrophe is incorrect. Trivial grammatical fix only.


johngibb:
@RomainMuller I know it's not an important change, but how do you get a review around here? :haha:

iliapolo (approved):
Thanks!

PR #2788, opened 2025-11-05
fix(import-schema): import PRs are not signed off and don't trigger the build workflow

leevs:
This would fix the DCO (signoff) correct labels and be in line with [dep upgrades](https://github.com/cdk8s-team/cdk8s/blob/master/.github/workflows/upgrade-dev-dependencies.yml)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2091,9 +2091,9 @@ base64-js@^1.3.1:
   integrity sha512-AKpaYlHn8t4SVbOHCy+b5+KKgvR4vrsD8vbvrbiQJps7fKDTkjkDry6ji0rUJjC0kzbNePLwzxq8iypo41qeWA==
 
 baseline-browser-mapping@^2.8.19:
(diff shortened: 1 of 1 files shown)


.github/workflows/import-k8s-schemas.yml
```diff
@@ -42,17 +42,24 @@ jobs:
...
+          token: ${{ secrets.PROJEN_GITHUB_TOKEN }}
           commit-message: >-
-            chore: import Kubernetes
+            chore(deps): import Kubernetes
             v${{ steps.check.outputs.latest }} schema
           branch: >-
             github-actions/k8s-schema-v${{ steps.check.outputs.latest }}
           title: >-
-            chore: import Kubernetes
+            chore(deps): import Kubernetes
```
iliapolo:
```suggestion
            chore: import Kubernetes
```

iliapolo (approved):

